PR #2, opened 2020-05-12
Add CurrentTenant class that can be resolved from container

AlexVanderbist:
Allows for things like this:

```php
use Spatie\Multitenancy\Models\Contracts\CurrentTenant;

Route::get('/current-tenant', function (CurrentTenant $currentTenant) {
    return $currentTenant; // returns the current Tenant model
});
```

### DIFF
--- a/docs/basic-usage/working-with-the-current-tenant.md
+++ b/docs/basic-usage/working-with-the-current-tenant.md
@@ -13,10 +13,18 @@ You can find the current method like this.
 Spatie\Multitenancy\Models\Tenant::current(); // returns the current tenant, or if not tenant is current, `null`
 ```
 
-A current tenant will also be bound in the container using the `currentTenant` key.
+A current tenant will also be bound in the container using the `currentTenant` key and `Spatie\Multitenancy\Models\Contracts\CurrentTenant` class.
 
 ```php
 app('currentTenant'); // returns the current tenant, or if not tenant is current, `null`
+```
+
+```php
+use Spatie\Multitenancy\Models\Contracts\CurrentTenant;
+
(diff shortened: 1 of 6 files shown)


docs/basic-usage/working-with-the-current-tenant.md
```diff
@@ -13,12 +13,20 @@ You can find the current method like this.
...
+A current tenant will also be bound in the container using the `currentTenant` key and `Spatie\Multitenancy\Models\Contracts\CurrentTenant` class.
 
 ```php
 app('currentTenant'); // returns the current tenant, or if not tenant is current, `null`
 ```
 
+```php
+use Spatie\Multitenancy\Models\Contracts\CurrentTenant;
+
+Route::get('/current-tenant', function (CurrentTenant $currentTenant) {
```
freekmurze, line 25: What happens if there is no current tenant?

AlexVanderbist: Good point, currently it'll new up a new `Tenant` instance but I'll change it to be `null`

AlexVanderbist: Hm, same issue exists for the `currentTenant`. I'll make both `null`

AlexVanderbist:
Too many issues: 
- wanted to throw an exception when resolving a currentTenant when there is no current tenant
- can't hook into resolving a key that doesn't exist in the container
- laravel will automatically create a new Tenant when trying to resolve `CurrentTenant::class`, even though it's literally bound to `null` 😠 

Gonna close this for now due to the above issues.

PR #3, opened 2020-05-12
Check a job payload has a tenantId index

themsaid:
In case jobs were in queue that don't have the tenantId index, it'll currently error. This check will prevent that.


themsaid:
Seem that it's already covered 👍🏼

PR #4, opened 2020-05-12
Allow multiple Tenant Domains

Rigby90:
First of all, appreciate the work on this and how quickly you've turned it from an idea into a solution!

I am building a solution for a client which does not require a tenant to have a domain, instead a user is able to select the current tenant by dropdown. This value is then stored in the session data and we set the correct tenant using a custom tenant finder. My only concern here is that currently a tenant 'requires' a domain name as the current table structure of the 'tenants' table forces a non-null unique value for the domains field. This is not a deal-breaker as we could create a random string and fill in the field for tenants which don't have a domain, and then find a way to filter it out from the backend interface when it's not a real domain.

In addition to this, there's also the requirement of a tenant having multiple domains. Whether this consists of the www and non-www variants of their domain, or additional domains that the client may hold and would like assigning to the same tenant.

With that being said, this PR should handle both situations. The PR is not complete in terms of tests or actual real-world testing, however, it provides the basis for a proof of concept and all tests pass. Only 1 existing test had to be modified slightly due to the domain being passed through to the Tenant to be created.

Would welcome any thoughts on this and what additional tests & logic may be required to get such a PR merged if at all possible.

Appreciate your time if you've made it this far 😄 

(PS, there's two additional commits to get the test suite working on Windows and to extract some additional MySQL details into the PHPUnit Environment Variables.)

### DIFF
--- a/config/multitenancy.php
+++ b/config/multitenancy.php
@@ -4,6 +4,7 @@ use Spatie\Multitenancy\Actions\ForgetCurrentTenantAction;
 use Spatie\Multitenancy\Actions\MakeQueueTenantAwareAction;
 use Spatie\Multitenancy\Actions\MakeTenantCurrentAction;
 use Spatie\Multitenancy\Actions\MigrateTenantAction;
+use Spatie\Multitenancy\Models\Domain;
 use Spatie\Multitenancy\Models\Tenant;
 use Spatie\Multitenancy\Tasks\PrefixCacheTask;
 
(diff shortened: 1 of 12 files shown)


freekmurze:
Thank you for your kinds words and your work on this PR.

I'm not going to merge it in, as I feel it goes against the philosophy of keeping the package as light as possible.

If you want to support multiple domains, use a custom `Tenant` model. In your custom tenant models you could add a relation to a domains table in your project. You can use a custom `TenantFinder` to look for tenants having a relation to the domain you want.


> PS, there's two additional commits to get the test suite working on Windows and to extract some additional MySQL details into the PHPUnit Environment Variables.

Could you send these changes in a separate PR? Thanks!

Rigby90:
Completely understand the reasoning, will proceed with a custom tenant model.

Will open another PR for the other two commits. Thanks.

PR #5, opened 2020-05-12
Support Testing on Windows & Extract config to PHPUnit Env

Rigby90:
As mentioned in #4 , a small change had to be made to support the migration imports during testing on Windows and in the process I extracted some additional variables to the env() helper to help with testing across different systems and users.


freekmurze:
Thank you!

PR #8, opened 2020-05-17
Asking for command and tenant

binaryk:
This was a very confusing for other packages. As how can we pass `--force` or any other custom options to command. 

image


freekmurze:
Thank you!

PR #9, opened 2020-05-17
Fix bug in landlord migration class name

SadeghPM:
This PR correct landlord migration class name and fix migration path.


freekmurze:
Thank you!

PR #11, opened 2020-05-17
Reconnect database connection

binaryk:


freekmurze:
I don't think this is needed.

@themsaid what's your thought on this?

themsaid:
The reconnection will happen next time your app requires a connection. If we connect now the request might not use the connection and it'd be a waste.

PR #14, opened 2020-05-19
Update using-multiple-databases.md

MichaelHappelOlsen:
Tenant::check() should be Tenant::checkCurrent() I guess?
and also a typo in seeder example php


freekmurze:
Thanks!

PR #16, opened 2020-05-19
One type and setting event's tenant prop to public

juukie:
Damn! Freek you're a fast merger ;-) Thanks. I've got a few extra changes.


freekmurze:
🚀

PR #17, opened 2020-05-21
Use the configured tenant model in artisan command

adriaanzon:
I have some custom methods on my tenant model which get called by a `SwitchTenantTask`. When running the `tenants:artisan` command, those methods didn't exist because it used `Spatie\Multitenancy\Models\Tenant` instead of  `App\Tenant` which I configured under `multitenancy.tenant_model`.


freekmurze:
Good catch! Thank you very much. I'll merge this is as soon as the tests of this PR finish.

freekmurze:
Thanks!

PR #18, opened 2020-05-21
No need to determine `$tenantId` twice...

telkins:
Minor change to DRY up the code a bit.  Hope I didn't miss anything.  😛

Also thought about moving the config check up above tenant ID check, but thought that micro-optimization might be a bit much.  😆

### DIFF
--- a/src/Actions/MakeQueueTenantAwareAction.php
+++ b/src/Actions/MakeQueueTenantAwareAction.php
@@ -49,7 +49,7 @@ class MakeQueueTenantAwareAction
                 return;
             }
             /** @var \Spatie\Multitenancy\Models\Tenant $tenant */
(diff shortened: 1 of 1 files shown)


freekmurze:
That's a nice change! Thanks!

telkins:
> That's a nice change! Thanks!

Np, @freekmurze .  Btw, can I ask how you figured out how to use `createPayloadUsing()` to add data to the payload?  The strategy you used is clean and simple, which is fantastic, but some of it's not documented.  I'm curious to hear if you learned of it from the source code...or some other "source".  🤓

freekmurze:
I learned this by watching @themsaid's videos on multitenancy. I've added a link in the credits section of the readme.

telkins:
Nice.  I've watched *some* of his videos.  Looks like I've missed at least one.  I'll have to go back to find out what I have missed.  🤓 

Thx.

PR #20, opened 2020-05-25
Remove unused import from config

Daanra:
Removes an unused import from the config.

(Or should it be registered under `'switch_tenant_tasks'`?)


freekmurze:
Thanks!

PR #21, opened 2020-05-26
Add down method to migration stub

CasperBE:
Make sure to drop the tenants table when resetting or refreshing the migrations.


freekmurze:
Thanks, but we are not using down migrations in any of our project/packages. We always migrate forwards.

Issue #43, opened 2020-06-11
Confusion about EnsureValidTenantSession


mpbarlow:
Hi. First off this is a great package, and apologies if I'm just misunderstanding something here. I notice that there is another open issue about this middleware, but I _think_ it's talking about a different issue, sorry if not.

[The documentation](https://docs.spatie.be/laravel-multitenancy/v1/installation/base-installation/#protecting-against-cross-tenant-abuse) says to register `EnsureValidTenantSession` as a global middleware. 

However if you do this, and your `TenantFinder` returns `null`, you get a `BindingResolutionException`, because `currentTenant` is never bound.

You can throw in your `TenantFinder`, but this won't work if you're using the provided `DomainTenantFinder`. This global middleware approach also prevents having any routes that are not tenant aware.

Providing I've not just made a stupid mistake somewhere, I've thought of a few potential solutions, and if the maintainers are on board with any I'd be happy to submit a PR:

- Modify `MultiTenantServiceProvider::determineCurrentTenant` to always bind `currentTenant`, even if it's `null`. I don't really like this approach, because the act of "making current" is all handled nicely by the Tenant model currently, and this would break that.

- Modify `EnsureValidTenantSession` such that it simply continues with the next request if no tenant is set. I don't _think_ this could cause any issues, as if there is no tenant I don't think there could be any tenant abuse? But it would be a breaking change -- same applies to updating `EnsureValidTenantSession` to throw `NoCurrentTenant` if the tenant is not bound.

- Update the documentation to instead recommend either: 
- Applying both `NeedsTenant` and `EnsureValidTenantSession` as global middleware, if all routes are tenant aware, or;
- Creating a new middleware group containing both `NeedsTenant` and `EnsureValidTenantSession`, if only some routes are tenant aware. This behaves as you would expect -- it first requires a tenant, throwing a nice easy-to-catch `NoCurrentTenant`, then validates the session to protect against abuse, giving a 401 if that check fails. 

I think the documentation option is probably the best as it requires no breaking changes (or any code changes at all)?

Please let me know what you think, like I say I'd be happy to open a PR for any of these.

freekmurze:
Send a PR for the documentation change 👍

PR #44, opened 2020-06-11
Clarify the use of EnsureValidTenantSession

mpbarlow:
As per #43, elaborate slightly on different approaches for global vs. route middleware to avoid exceptions when using `EnsureValidTenantSession` globally.


freekmurze:
Thanks!

PR #46, opened 2020-06-12
tenants:artisan search tenant by ids or domains

masterix21:
Ids are great, but isn't more simple to use tenants:run also by domains?

Search by ID:
`php artisan tenants:artisan route:list --tenant=1`

Search by Domain:
`php artisan tenants:artisan route:list --tenant=tenant1.spatie.be`

Mixed search:
`php artisan tenants:artisan route:list --tenant=2,tenant1.spatie.be`


freekmurze:
Thanks for your work on this. It's a nice suggestion, but it wouldn't work if users remove the `domain`attribute. It is far less likely that the `id` parameter is going to be removed.

Keeping the philosophy of this package in mind (keeping everything as light as it can be), I'm going to pass on this PR for now.

PR #47, opened 2020-06-14
Unset the previous tenant if a queue job cannot load its own tenant

mpbarlow:
This is a fix for behaviour that I believe could lead to unintended data disclosure if you have a mixture of tenant and non-tenant routes in your app, or if a tenant is deleted before its queue job runs. 

The problem lies in the fact that each queue worker only boots the app once—any global state changes made by a queue job will still be present for the next queue job. 

If a tenant aware job cannot fetch its tenant (either because it no longer exists or because it was not set when dispatched), that job will continue to run under whatever tenant the last job used.

As the simplest example, imagine the following routes and job, running on a Redis queue.

```php
Route::get('tenant', function () {
    Tenant::first()->makeCurrent();

    dispatch(new TestJob());

    return Tenant::checkCurrent() ? 'active tenant' : 'no tenant';
});

Route::get('no-tenant', function () {
    dispatch(new TestJob());

    return Tenant::checkCurrent() ? 'active tenant' : 'no tenant';
});

class TestJob implements ShouldQueue, TenantAware
{
    public function handle()
    {
        echo Tenant::checkCurrent() ? 'active tenant' : 'no tenant';
    }
}
```

Visiting `/tenant` displays 'active tenant' and visiting `/no-tenant` displays 'no tenant', as you would expect. However, looking at the output of the queue worker, you will see `active tenant` followed by `active tenant` again.

Obviously this example is fairly contrived, but a queue job that sends emails or updates records could cause some trouble.

The simplest option would be listening for `JobProcessed` and forgetting the tenant there, but that causes problems when using the sync queue driver (as it forgets the tenant in the middle of your code). 

As such I have made the following changes:
* I have modified the payload callback very slightly. If the job is _not_ tenant aware, no `tenantId` key is added (exactly as before). However, if the job _is_ tenant aware, a `tenantId` key is _always_ added, though it may be null. When coming to retrieve the tenant when the job is popped from the queue, this makes it possible to tell the difference between jobs that are not tenant aware, and those that are, but where no tenant was set at dispatch time.
* When the job is processed, if there is a `tenantId` present in the payload (i.e. it is tenant aware), but that `tenantId` is null or non-existent, the active tenant is unset. This will prevent tenant aware jobs from accidentally running under a previous tenant.
* If there is no `tenantId` field at all, nothing happens. The active tenant is left as it was. I think this should be okay, as a non-tenant job should never be making any assumptions about the tenant anyway.

As an aside, I have also removed the check that returns early if `queues_are_tenant_aware_by_default` is false, as that appeared to be preventing the tenant being set even if the job implemented `TenantAware` (`$tenant->makeCurrent();` becomes an unreachable statement).

I don't _think_ this is a breaking change, the only behaviour that is different is that `tenantId` on the payload is now sometimes null instead of missing. 

I'm happy to make any changes if the maintainers would like any.

### DIFF
--- a/src/Actions/MakeQueueTenantAwareAction.php
+++ b/src/Actions/MakeQueueTenantAwareAction.php
@@ -3,6 +3,7 @@
 namespace Spatie\Multitenancy\Actions;
 
 use Illuminate\Queue\Events\JobProcessing;
+use Spatie\Multitenancy\Exceptions\NoCurrentTenant;
 use Spatie\Multitenancy\Jobs\NotTenantAware;
 use Spatie\Multitenancy\Jobs\TenantAware;
 use Spatie\Multitenancy\Models\Concerns\UsesTenantModel;
@@ -28,9 +29,7 @@ class MakeQueueTenantAwareAction
                 return [];
             }
 
-            return Tenant::current()
-                ? ['tenantId' => Tenant::current()->id]
-                : [];
+            return ['tenantId' => optional(Tenant::current())->id];
         });
 
         return $this;
@@ -39,18 +38,15 @@ class MakeQueueTenantAwareAction
     protected function listenForJobsBeingProcessed(): self
     {
         app('events')->listen(JobProcessing::class, function (JobProcessing $event) {
-            $tenantId = $event->job->payload()['tenantId'] ?? null;
-
-            if (! $tenantId) {
+            if (! array_key_exists('tenantId', $event->job->payload())) {
                 return;
             }
 
-            if (! config('multitenancy.queues_are_tenant_aware_by_default')) {
-                return;
-            }
+            $tenantId = $event->job->payload()['tenantId'];
+
             /** @var \Spatie\Multitenancy\Models\Tenant $tenant */
-            if (! $tenant = $this->getTenantModel()::find($tenantId)) {
-                return;
+            if (! $tenantId || ! ($tenant = $this->getTenantModel()::find($tenantId))) {
+                throw NoCurrentTenant::make();
             }
 
             $tenant->makeCurrent();
--- a/tests/Feature/TenantAwareJobs/JobIsDeletedWhenTenantCannotBeRetrievedTest.php
+++ b/tests/Feature/TenantAwareJobs/JobIsDeletedWhenTenantCannotBeRetrievedTest.php
@@ -1,0 +1,110 @@
+<?php
+
+
+namespace Spatie\Multitenancy\Tests\Feature\TenantAwareJobs;
+
+use Illuminate\Contracts\Bus\Dispatcher;
+use Spatie\Multitenancy\Exceptions\NoCurrentTenant;
+use Spatie\Multitenancy\Models\Tenant;
+use Spatie\Multitenancy\Tests\Feature\TenantAwareJobs\TestClasses\NotTenantAwareTestJob;
+use Spatie\Multitenancy\Tests\Feature\TenantAwareJobs\TestClasses\TenantAwareTestJob;
+use Spatie\Multitenancy\Tests\Feature\TenantAwareJobs\TestClasses\TestJob;
+use Spatie\Multitenancy\Tests\TestCase;
+use Spatie\Valuestore\Valuestore;
+
+class JobIsDeletedWhenTenantCannotBeRetrievedTest extends TestCase
+{
+    private Tenant $tenant;
+
+    private Valuestore $valuestore;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+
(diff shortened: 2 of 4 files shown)


freekmurze:
Thank for your work on this. I think this solve a valid issue. I'd like it to be handle differently though.

Instead of merely unsetting the active tenant and running the job using landlord configuration, I think we should outright delete the job. It would work a bit like [Laravel's missing models queue feature](https://laravel.com/docs/7.x/queues#ignoring-missing-models). So in our case, the job should be deleted if the job was tenant aware and the tenant doesn't exist anymore.

Could you make that change to this PR?

mpbarlow:
This turned out to be a surprisingly difficult problem to solve!

Somewhat unexpectedly, deleting or failing a job in a `JobProcessing` callback does not actually prevent execution of the job when using the sync driver (although it does appear to for the redis and database drivers), so that was a non-starter.

However, throwing during the callback ticks the following boxes:
- The job's `handle` method is not called
- The job's `failed` method _is_ called (with an instance of `NoCurrentTenant`), which could be useful for logging or taking some other actions
- The job is _not_ retried, which is good because if it failed due to a missing tenant, it's never going to succeed
- Works as expected under the sync, redis, and database queue drivers (I have no provisions for testing SQS)

I'm not sure I totally love the solution—I'd feel better if there was some cancel API for jobs—but unless I'm missing something obvious this is the only way I've found that works in all situations.

I've also added a small update to the docs to cover this behaviour. Thanks.

src/Actions/MakeQueueTenantAwareAction.php
```diff
@@ -39,18 +38,15 @@ protected function listenForJobsBeingQueued(): self
...
 
-            if (! config('multitenancy.queues_are_tenant_aware_by_default')) {
-                return;
-            }
+            $tenantId = $event->job->payload()['tenantId'];
+
             /** @var \Spatie\Multitenancy\Models\Tenant $tenant */
-            if (! $tenant = $this->getTenantModel()::find($tenantId)) {
-                return;
+            if (! $tenantId || ! ($tenant = $this->getTenantModel()::find($tenantId))) {
```
freekmurze, line 48: Instead of using a compound statement here, let's extra the check to a dedicated function where we use early returns.

tests/Feature/TenantAwareJobs/JobIsDeletedWhenTenantCannotBeRetrievedTest.php
```diff
@@ -0,0 +1,110 @@
...
+use Illuminate\Contracts\Bus\Dispatcher;
+use Spatie\Multitenancy\Exceptions\NoCurrentTenant;
+use Spatie\Multitenancy\Models\Tenant;
+use Spatie\Multitenancy\Tests\Feature\TenantAwareJobs\TestClasses\NotTenantAwareTestJob;
+use Spatie\Multitenancy\Tests\Feature\TenantAwareJobs\TestClasses\TenantAwareTestJob;
+use Spatie\Multitenancy\Tests\Feature\TenantAwareJobs\TestClasses\TestJob;
+use Spatie\Multitenancy\Tests\TestCase;
+use Spatie\Valuestore\Valuestore;
+
+class JobIsDeletedWhenTenantCannotBeRetrievedTest extends TestCase
```
freekmurze, line 15: I appreciate the added tests 👍

tests/Feature/TenantAwareJobs/JobIsDeletedWhenTenantCannotBeRetrievedTest.php
```diff
@@ -0,0 +1,110 @@
...
+    /** @test */
+    public function it_will_fail_a_job_when_no_tenant_is_present_and_queues_are_tenant_aware_by_default()
+    {
+        config()->set('multitenancy.queues_are_tenant_aware_by_default', true);
+
+        $job = new TestJob($this->valuestore);
+
+        try {
+            app(Dispatcher::class)->dispatch($job);
+        } catch (NoCurrentTenant $e) {
```
freekmurze, line 42: In most cases, I don't use abbreviations. Please use `$exception` instead of `$e`

src/Actions/MakeQueueTenantAwareAction.php
```diff
@@ -39,18 +38,15 @@ protected function listenForJobsBeingQueued(): self
...
+                throw NoCurrentTenant::make();
```
freekmurze, line 49: Instead of using this exception, let's create a dedicated exception for this situation. Might be called `NoCurrentTenantFoundInTenantAwareJob`

freekmurze:
Thanks! I'll polish this one further myself

PR #52, opened 2020-06-18
tenant:artisan search customizable using config

masterix21:
Hi, 

i know that isn't a priority to use "tenant:artisan" searching by domain, but now I think it could be merged.

Now it works by id (like past), or by your custom search fields from config file:
```php
'tenant_artisan_search_fields' => 'domain', // will search by domain
// OR
'tenant_artisan_search_fields' => [ 'id', 'domain' ], // will search by id or domain
// ...
```

If you identify your tenant using a custom field, it will work too:
```php
// ...
'tenant_artisan_search_fields' => 'host', // will search by any other column in tenant model
// ...
```

### DIFF
--- a/config/multitenancy.php
+++ b/config/multitenancy.php
@@ -18,10 +18,10 @@ return [
 
     /*
      * These fields are used by tenant:artisan command to match one or more tenant
-     *
-     * Set to `null` to use only id
      */
-    'tenant_artisan_search_fields' => null,
+    'tenant_artisan_search_fields' => [
+        'id',
+    ],
 
     /*
      * These tasks will be performed when switching tenants.
--- a/config/multitenancy.php
+++ b/config/multitenancy.php
@@ -17,6 +17,13 @@ return [
     'tenant_finder' => null,
 
     /*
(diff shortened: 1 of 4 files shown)


config/multitenancy.php
```diff
@@ -16,6 +16,13 @@
      */
     'tenant_finder' => null,
 
+    /*
+     * These fields are used by tenant:artisan command to match one or more tenant
+     *
+     * Set to `null` to use only id
```
freekmurze, line 22: Let's set it to `id` by default

src/Concerns/UsesMultitenancyConfig.php
```diff
@@ -35,4 +36,9 @@ public function getMultitenancyActionClass(string $actionName, string $actionCla
 
         return app($configuredClass);
     }
+
+    public function getTenantArtisanSearchFields() : array
+    {
+        return Arr::wrap(config('multitenancy.tenant_artisan_search_fields') ?? 'id');
```
freekmurze, line 42: Using `id` instead of `null` isn't necessary when `id` is set as the default in the config file.

freekmurze:
Thanks!

PR #57, opened 2020-06-21
allow mass assignment when creating a new tenant by default

eightfivelabs:
Hi,

It seems by default that the below doesn't work, should this be changed?

```php
$tenant = Tenant::create([
  'name' => 'tenantX-name',
  'domain' => 'tenantX-domain',
  'database' => 'tenantX-database'
]);
```


freekmurze:
Thanks!

Issue #51, opened 2020-06-17
Sanctum package not switching the database connection


redonosmanollaj:
I am using sanctum package for authorizing users in both tenant database connection and landlord database connection. In my case I'm using example.test domain for authorizing landlord users, and subdomains ex. subdomain.example.test for authorizing tenant users. I am using landlord connection as default and works good when trying to login as landlord user, but when trying to login in subdomain.example.test it looks like sanctum cannot change database connection to 'tenant'.

masterix21:
Hi @redonosmanollaj, 

are you using the SwitchTenantDatabaseTask class?

I have tried now and it seems ok:

<img width="1678" alt="tinker" src="https://user-images.githubusercontent.com/6555012/84955300-53f1fd00-b0f7-11ea-8345-faec93e9be6e.png">

redonosmanollaj:
> Hi @redonosmanollaj,
> 
> are you using the SwitchTenantDatabaseTask class?
> 
> I have tried now and it seems ok:
> 
> <img alt="tinker" width="1678" src="https://user-images.githubusercontent.com/6555012/84955300-53f1fd00-b0f7-11ea-8345-faec93e9be6e.png">

Yes tokens generated good, but when trying to call routes that have sanctum middleware, landlord works well, but tenantToken doesnt work because sanctum is trying to find token in landlord database.

redonosmanollaj:
Or vice versa, if I set tenant as default connection, tenant routes works well, but when I try to call landlord routes I get this error:

Illuminate\Database\QueryException: SQLSTATE[3D000]: Invalid catalog name: 1046 No database selected (SQL: select * from `personal_access_tokens` where `personal_access_tokens`.`id` = 1 limit 1) in file C:\laragon\www\strike-backend\vendor\laravel\framework\src\Illuminate\Database\Connection.php on line 671

#0 C:\laragon\www\strike-backend\vendor\laravel\framework\src\Illuminate\Database\Connection.php(631): Illuminate\Database\Connection->runQueryCallback('select * from `...', Array, Object(Closure))
#1 C:\laragon\www\strike-backend\vendor\laravel\framework\src\Illuminate\Database\Connection.php(339): Illuminate\Database\Connection->run('select * from `...', Array, Object(Closure))
#2 C:\laragon\www\strike-backend\vendor\laravel\framework\src\Illuminate\Database\Query\Builder.php(2202): Illuminate\Database\Connection->select('select * from `...', Array, true)
[...]

Issue #29, opened 2020-05-29
Multitenancy with Passport, Scheduling, Filesystem


anwardote:
Hi,

Great work! I am really very excited to use this package at my running project.

I have read DOCs. I could not find any settings regarding API Passport and scheduling. Doesn't have any specific Settings for managing API requests and managing scheduling jobs?

If I want to use a different File storage directory for each Tenant managing File storage, Does there have any method/settings to maintain this?

Anwar

freekmurze:
No, the package does not do this. It allows you to define tasks that should run when switching tenants. If you expect a more batteries included type of package, please check out the alternatives mentioned in the readme.

anwardote:
Thank you for your quick response.

However, As I research the packages DOCs, I got well flexibility. So, I will recommend this package for my project. In this case, I need your help to resolve those issues.

**FileSystem:** For these issues, I can create a new Task for managing the File System easily. 
In the makeCurrent, I can write a few code as follows:
`config()->set('filesystems.disks.disk_name.root', storage_path("app/public/tenant_{$tenant->id}" ));`
`config()->set('filesystems.disks.disk_name.url', "/storage/tenant_{$tenant->id}");`
  
**Passport:**  I saw a reference to resolve this.(Not test in this package). 
[https://medium.com/@wannymiarelli/laravel-multi-tenancy-with-passport-10ce9b6a1c21](https://medium.com/@wannymiarelli/laravel-multi-tenancy-with-passport-10ce9b6a1c21
)

**Scheduling:** I am really worried for scheduling matters. In my project, might have 5000+ Tenants in the future. Can you provide any hints for that?

I highly request you to review my above possible solution and suggest me the best way.

Thank you in advance.
Anwar

masterix21:
@anwardote, it's effortless using tasks. 

You can add a "tenant" disk in "config/filesystems.php" file:

```php
'disks' => [
      'tenant' => [
            'visibility' => 'public',
      ],
]
```

Next, you can create your switch task "FilesystemTask":
```php
class FilesystemTask implements SwitchTenantTask
{
    public function makeCurrent(Tenant $tenant): void
    {
        $this->setTenantFilesystem($tenant);
    }

    public function forgetCurrent(): void
    {
        $this->setTenantFilesystem();
    }

    protected function setTenantFilesystem(?Tenant $tenant = null): void
    {
        config()->set('filesystem.tenant', [
                // your setup for a public disk
        ]);
    }
}
```

[...]

PR #59, opened 2020-06-21
Change database default connection if needed

masterix21:
Using packages like Laravel Passport or Laravel Sanctum (and many others) could be useful to switch the default database connection to "tenant" instead of "landlord".

Using the config file, now it is merely possible:

```php
/*
 * Using `\Spatie\Multitenancy\Tasks\SwitchTenantDatabaseTask`, could be useful to
 * set up the tenant database connection as default connection.
 */
'tenant_database_connection_as_default' => true,
```

### DIFF
--- a/config/multitenancy.php
+++ b/config/multitenancy.php
@@ -59,6 +59,12 @@ return [
     'landlord_database_connection_name' => null,
 
     /*
+     * Using `\Spatie\Multitenancy\Tasks\SwitchTenantDatabaseTask`, could be useful to
+     * set up the tenant database connection as default connection.
+     */
+    'tenant_database_connection_as_default' => false,
+
+    /*
      * This key will be used to bind the current tenant in the container.
      */
     'current_tenant_container_key' => 'currentTenant',
--- a/src/Concerns/UsesMultitenancyConfig.php
+++ b/src/Concerns/UsesMultitenancyConfig.php
@@ -17,6 +17,11 @@ trait UsesMultitenancyConfig
         return config('multitenancy.landlord_database_connection_name') ?? config('database.default');
     }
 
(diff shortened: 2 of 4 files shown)


freekmurze:
Couldn't you just set a tenant database as the default in the. `databases.php` config file if you'd like that?

masterix21:
Your reasoning is not wrong, but there seem to be problems. Following the past issues, appears to be a solution to the questions:
- https://github.com/spatie/laravel-multitenancy/issues/51
- https://github.com/spatie/laravel-multitenancy/issues/29#issuecomment-636499851

I tried and seemed necessary. Furthermore, could be a way to skip the "--database=" argument with migrations, or other commands where the "--database=" argument is unimplemented (and there are a lot of packages).

freekmurze:
> Your reasoning is not wrong, but there seem to be problems.

I don't fully understand the problem or how this PR solves the problem, but adding an extra configuration option for something that can already be done in configuration seems off.

I'd prefer a solution where no config option is added. Maybe the current implementation can be changed in a non-breaking way that solves the problem?

masterix21:
See the attached video where I tried to explain the problem using Tinkerwell. I don't know if there are other solutions, but changing the default database connection works.

I will inspect better in the problem, but if you have an idea or understand what should be, please keep me updated.

example

masterix21:
Another example.

It happens when you use a model without UsesTenantConnection (or UsesLandlordConnection).

example

masterix21:
A partial solution (already tested) could be a trait with:

```php
public function getConnectionName()
{
        return Tenant::checkCurrent()
            ? 'tenant'
            : 'landlord';
}
```

But it means that you must extend all original models with new ones that use the trait.

Issue #69, opened 2020-06-30
tenant migration adds to landlord database


aminraeisi:
Hi,

Thanks for the great package!

I just started testing the package and created some records directly into the `landlord.tenants` table and ran `php artisan tenants:artisan migrate` but it adds them to my landlord database (only once).

here is my files and the settings I am using:

```php

// config/database.php

return [

    'default' => env('DB_CONNECTION', 'landlord'),


    'connections' => [

        'tenant' => [
            'driver' => 'mysql',
            'host' => env('DB_HOST', '127.0.0.1'),
            'port' => env('DB_PORT', '3306'),
            'database' => null,
            'username' => env('DB_USERNAME', 'forge'),
            'password' => env('DB_PASSWORD', ''),
            'unix_socket' => env('DB_SOCKET', ''),
            'charset' => 'utf8mb4',
            'collation' => 'utf8mb4_unicode_ci',
            'prefix' => '',
            'prefix_indexes' => true,
            'strict' => false,
            'engine' => null,
            'options' => extension_loaded('pdo_mysql') ? array_filter([
                PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),
            ]) : [],
        ],

        'landlord' => [
            'driver' => 'mysql',
            'host' => env('DB_HOST', '127.0.0.1'),
            'port' => env('DB_PORT', '3306'),
            'database' => 'landlord',
            'username' => env('DB_USERNAME', 'forge'),
            'password' => env('DB_PASSWORD', ''),
            'unix_socket' => env('DB_SOCKET', ''),
            'charset' => 'utf8mb4',
            'collation' => 'utf8mb4_unicode_ci',
            'prefix' => '',
            'prefix_indexes' => true,
            'strict' => false,
            'engine' => null,
            'options' => extension_loaded('pdo_mysql') ? array_filter([
                PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),
            ]) : [],
        ],
]

```

```php

// config/multitenancy.php

return [

    'tenant_finder' => DomainTenantFinder::class,

    'tenant_artisan_search_fields' => [
        'id',
    ],

    'switch_tenant_tasks' => [
        Spatie\Multitenancy\Tasks\SwitchTenantDatabaseTask::class,
    ],


    'tenant_model' => Tenant::class,

    'queues_are_tenant_aware_by_default' => true,

[...]

PR #72, opened 2020-07-01
Fix database migration doc

masterix21:
Added `--database=` argument to `migrate` command in the documentation.

It will help the users to prevent the issue https://github.com/spatie/laravel-multitenancy/issues/69#issuecomment-652672865


freekmurze:
Thank you!

Issue #90, opened 2020-07-21
SwitchTenantDatabaseTask Issue with UsesLandlordConnection


medo3337:
When using SwitchTenantDatabaseTask, UsesLandlordConnection models are trying to mistakenly connect to tenant instead of landlord database.

If I remove SwitchTenantDatabaseTask, UsesLandlordConnection models seems to be making valid connection to the landlord database.

freekmurze:
I'd appreciate a PR that adds a fix and a tests for this that makes sure that it works.

masterix21:
@medo3337 are you using a custom `SwitchTenantDatabaseTask`?

To simulate your environment, I have created two models:
```php
class LandlordUser extends User
{
    use UsesLandlordConnection;

    protected $table = 'users';
}

class TenantUser extends User
{
    use UsesTenantConnection;

    protected $table = 'users';
}
```

And using Tinkerwell:
![image](https://user-images.githubusercontent.com/6555012/88198993-99e53800-cc44-11ea-8c69-2ac7dbd612a4.png)

Everythings seems that works well.

medo3337:
@masterix21 I was able to fix this issue by setting tenant_database_connection_name and landlord_database_connection_name config variables.

If you want to reproduce, try setting both variables to null. I assume that both landlord and tenant was trying to reach for default connection and hence both were connecting to tenant.

Maybe exception should be thrown if any config variable is required?

masterix21:
@freekmurze is there a reason to set `tenant_database_connection_name` to `null` using `SwitchTenantDatabaseTask`? Or can I throw the exception with a PR?

freekmurze:
> is there a reason to set tenant_database_connection_name to null using SwitchTenantDatabaseTask

I'm thinking it's the save thing to do that if you want to forget the current tenant, the connection to its database should not be set anymore.

masterix21:
My interpretation of `SwitchTenantDatabaseTask` is switch your default connection to another one: it means that `tenant_database_connection_name` can't be `null`.

A `null` value for it, is allowed, but only when you don't need to switch the connection (it's the case of a one-database app).

See it: https://github.com/spatie/laravel-multitenancy/pull/92

PR #92, opened 2020-07-24
Database switch fails with misconfigured tenant

masterix21:
Using `SwitchTenantDatabaseTask` it's better to have two different connection names between landlord and tenant.

To avoid a misconfiguration, I have added an exception throws when tenant `multitenancy.tenant_database_connection_name` is the same of `multitenancy.landlord_database_connection_name`.

Related issue: https://github.com/spatie/laravel-multitenancy/issues/90


freekmurze:
Thanks!

PR #93, opened 2020-07-24
:cookie: change the failed job database in 'SwitchTenantDatabaseTask'

### LINKED ISSUE #91: when there's a `Failed tenant job` it will be inserted in landlord database
there's two database `for tenant` and `for landlord`

#### Event:
``
public function __construct(TestModel $model)
{
    $this->model = $model;
}
``

#### Listener function:
`
public function handle(TestEvent $event): void
    {
        throw new \Exception("Test record failed jobs");
    }
`

-> database of landlord has failed_jobs table
-> database of tenant has failed_jobs table

### result:
in the tenant failed_jobs table is empty
in the landlord failed_jobs table is:
`
Exception: Test record failed jobs in D:\fullpath\ListenerFile.php:53
Stack trace:
...
`

abderrazzak-oxa:
Fixes #91


masterix21:
@abderrazzak-oxa, it won't works to who centralize all jobs in the landlord connection also using the `SwitchTenantDatabaseTask`. 

Should be better to create a boolean parameter in the configuration file. Don't you agree with me?

abderrazzak-oxa:
Yes but the `config.queue.failed.database` must be changed when the there's a change from landlord to tenant or vice versa

masterix21:
If I understood your problem, you like to migrate `jobs` and `failed_jobs` in each tenant database. With your changes, if somebody wants to migrate `jobs` and `failed_jobs` in the landlord database, keeping your all jobs tenant-aware, could not.

I think could be better to create a configuration to allow everyone to make their choices, or - better - a `SwitchTenantJobsDatabaseTask`: it will keep fully customizable the package.

abderrazzak-oxa:
or we can do this when the tenant is forgot we can assign the old database connection to failed.database like this:

```
class SwitchTenantDatabaseTask implements SwitchTenantTask
{
    use UsesMultitenancyConfig;

    protected $failedJobDatabaseConnection;

    public function __construct()
    {
        $this->failedJobDatabaseConnection = config('queue.failed.database');
    }

    public function makeCurrent(Tenant $tenant): void
    {
        $this->setTenantConnectionDatabaseName($tenant->getDatabaseName());
    }

    public function forgetCurrent(): void
    {
        $this->setTenantConnectionDatabaseName(null);
    }

    protected function setTenantConnectionDatabaseName(?string $databaseName): void
    {
        $tenantConnectionName = $this->tenantDatabaseConnectionName();

        if ($tenantConnectionName === $this->landlordDatabaseConnectionName()) {
            throw InvalidConfiguration::tenantConnectionIsEmptyOrEqualsToLandlordConnection();
        }

        if (is_null(config("database.connections.{$tenantConnectionName}"))) {
            throw InvalidConfiguration::tenantConnectionDoesNotExist($tenantConnectionName);
        }

        config([
            "database.connections.{$tenantConnectionName}.database" => $databaseName,
            "queue.failed.database" => $databaseName === null ? $this->failedJobDatabaseConnection : $tenantConnectionName
        ]);

        DB::purge($tenantConnectionName);
    }
}
```

are you agree with me?

abderrazzak-oxa:
Yes i think it's best to create `SwitchTenantJobsDatabaseTask`, and anyone can create a new tenant task.

but the tenant should change the job database by default or we can create new task called `SwitchTenantJobsDatabaseTask`

masterix21:
I think you can create `SwitchTenantJobsDatabaseTask`

abderrazzak-oxa:
@masterix21 it's created!

masterix21:
Failed jobs are now ok, but what about the jobs table? Don't you think is better to change `queue.connections.database.connection` also? 

Finally, please add the tests.

abderrazzak-oxa:
i think you can't change the queue connection in laravel, if you want to change the queue connection you need to restart this command: `php artisan queue`.

so if you use database connection in the queue you must use one database connection.

so the jobs table must be in one database, i am using redis and the failed job table is in tenant databases and this task is works fine

masterix21:
Why not running `php artisan tenants:artisan "queue:listen" --tenant=xxx` (for each tenant)? If you have a load balancer it could be useful to who can't use Redis.

abderrazzak-oxa:
Ok so what if there's new tenant how can you execute this command, and when you use `supervisor` in `ubuntu`? 

you need to go to `supervisor` config and add the new command for tenant like: `php artisan tenants:artisan "queue:listen" --tenant=2`

masterix21:
I don't know your setup, but from Laravel Forge (for example), you could use the API to set up your new one "queue:listen". 

Anyway, it's better to wait for @freekmurze's thinks. Honestly, I'd prefer to keep the package clean and without any additional jobs setup, but I'm not the one who decides.

Finally, please remember the tests: without, it's harder that @freekmurze will accept the PR.

abderrazzak-oxa:
Yes i know when you want to start the jobs you must use this command for each tenant `php artisan tenants:artisan "queue:listen" --tenant=$tenantID`

but there's no config in this path: `queue.connections.database.connection` to change the database!

#### queue config is:
```
'database' => [
            'driver' => 'database',
            'table' => 'jobs',
            'queue' => 'default',
            'retry_after' => 90,
        ],
```

masterix21:
https://stackoverflow.com/questions/35535920/laravel-5-how-to-configure-the-queue-database-driver-to-connect-to-a-non-default

https://stackoverflow.com/questions/41318374/running-laravel-queue-with-multiple-database

abderrazzak-oxa:
Ok so it's good,

now feel free to decline or accept 

@freekmurze

freekmurze:
@abderrazzak-oxa Thanks for your work on this. Could you add some tests and update the docs? Thanks!

abderrazzak-oxa:
@freekmurze, Yes i will do that.

abderrazzak-oxa:
@freekmurze @masterix21  The test is failed ! why?

masterix21:
```
1) Spatie\Multitenancy\Tasks\SwitchTenantJobsDatabaseTest::when_making_a_tenant_current_the_jobs_will_stored_in_his_database
Failed asserting that false is true.

/home/runner/work/laravel-multitenancy/laravel-multitenancy/tests/Feature/Tasks/SwitchTenantJobsDatabaseTest.php:40
```

abderrazzak-oxa:
Hello @masterix21 ,
please check now the problem i found is when change the config `queue.connections.database.connection`
it can't be reconnect!
so how can we reconnect the queue database,

masterix21:
Do you have the code to catch the problem?

abderrazzak-oxa:
Let me to explain to you,

#### this is the tests
// change tenant and change the queue connection database `queue.connections.database.connection`
// dispatch job
$hasJobs = DB::table('jobs')->exists();
$this->assertTrue($hasJobs);

so this test is failed because the `queue.connections.database.connection` is changed but it's not reconnected,


because the laravel just can connect queue one time when the laravel app is started:

#### vendor\laravel\framework\src\Illuminate\Queue\Connectors\DatabaseConnector.php 
```
/**
     * Establish a queue connection.
     *
     * @param  array  $config
     * @return \Illuminate\Contracts\Queue\Queue
     */
    public function connect(array $config)
    {
        return new DatabaseQueue(
            $this->connections->connection($config['connection'] ?? null),
            $config['table'],
            $config['queue'],
            $config['retry_after'] ?? 60
        );
    }
```

#### and i don't found any way to reconnect the queue when tenant changed

- note: The failed jobs is works fine

Issue #74, opened 2020-07-06
Testing Issue


sausin:
Hi!

Thanks for the wonderful work on this package! The integration with the app in progress went quite well and things are working well on the actual app (I think!).

One snag that I seem to have hit is with testing. All the tests were working correctly before integration with this package. I made the following changes to get things working:

- Added the following method to the `TestCase.php` file

```php
public function setUp(): void
{
    parent::setUp();

    Tenant::find(1)->makeCurrent();
}
```

- Updated the `multitenancy.php` config file so it would have the test database information for `landlord` and `tenant`
```php
'tenant_database_connection_name' => env('DB_CONNECTION_TENANT', null),
'landlord_database_connection_name' => env('DB_CONNECTION_LANDLORD', null),
```
- The `.env.testing` file defines these variables with information for the databases and the same connections are defined in the `database.php` config file

Currently, most of the tests are working, but some are throwing up random errors. The thing that is common in all tests where failures are occurring is the use of `DatabaseTransactions` trait.

Checking further, I found that the trait is infact not working. If some entries are created / deleted in the tests, they do not get reverted.

Any help would be great to have :-)

masterix21:
hi @sausin, can you post a test example, please?

sausin:
Sure! Please see below an example test on the reset route:

```php
<?php

namespace Tests\Feature\Controllers\Auth;

use App\User;
use Illuminate\Auth\Notifications\ResetPassword;
use Illuminate\Foundation\Testing\DatabaseTransactions;
use Illuminate\Support\Facades\Event;
use Illuminate\Support\Facades\Hash;
use Illuminate\Support\Facades\Notification;
use Illuminate\Support\Str;
use Tests\TestCase;

class ForgotPasswordControllerTest extends TestCase
{
    use DatabaseTransactions;

    /** @test */
    public function it_resends_verification_email()
    {
        Event::fake();
        Notification::fake();

        $u = User::create([
            'name' => Str::random(10),
            'email' => $e = Str::random(10).'@example.com',
            'phone_number' => 00001112345,
            'password' => Hash::make('password'),
        ]);

        $this->postJson('/password/email', ['email' => $e])
            ->assertOk();

        Notification::assertSentTo($u, ResetPassword::class);
    }
}
```

[...]

PR #101, opened 2020-07-31
Testing setup when using DatabaseTransactions trait

sausin:
Docs update as per discussion and resolution in #74


freekmurze:
Thanks!

Issue #103, opened 2020-08-04
Using Laravel Scheduler with Multitenancy


LukeAbell:
Running `php artisan tenants:artisan schedule:run` loops through the tenants and runs the scheduled commands but Tenant::current() is null. I think I've narrowed this down to the scheduler calling an artisan command in a forked process which recalls `php artisan XXX` with whatever the command is (which bypasses the multi-tenancy call). 

Is there a way to make the scheduler tenant aware like the package currently does with queues?

Thanks!

masterix21:
Hi @LukeAbell, are you sure that are you doing everything well?

See my example here:
https://github.com/spatie/laravel-multitenancy/issues/82#issuecomment-663771093

Please post your code. Thank you

LukeAbell:
@masterix21 @freekmurze The issue only occurs when using `$schedule->command()` or `$schedule->exec()`. Running `$schedule->job()` on the works correctly.

LukeAbell:
@freekmurze The issue also affects all of the Mailcoach commands. Turning them into Jobs works.

masterix21:
@LukeAbell, thanks for your issue. 

I'm working over the solution, and in the next days, I will release the PR to fix it.

In the meanwhile, if you like to apply a temporary solution, you can use this trait with your tenant-aware commands:
```php
trait IsTenantAware
{
    protected function execute(InputInterface $input, OutputInterface $output)
    {
        $tenantIds = Arr::wrap($this->option('tenant'));

        return Tenant::query()
            ->when(! blank($tenantIds), fn ($query) => $query->whereIn('id', $tenantIds))
            ->cursor()
            ->map(fn ($tenant) => $tenant->execute(fn () => (int) $this->laravel->call([$this, 'handle'])))
            ->sum();
    }
}
```

Finally, change your command signatures appending `--tenant=*`.

LukeAbell:
@masterix21 thank you!!

masterix21:
I forgot to say that you could change your schedule method as follow:
```php
    protected function schedule(Schedule $schedule)
    {
        $schedule->command('tenant:aware:command');
    }
```

And you could call only
```bash
php artisan schedule:run
```

Thanks

masterix21:
@LukeAbell the fix has been released.

LukeAbell:
@masterix21 thank you!!

PR #106, opened 2020-08-05
Feature command tenant aware

masterix21:
Lightening fast solution to the issue https://github.com/spatie/laravel-multitenancy/issues/103


freekmurze:
That's a good solution, thank you very much!

I'll polish this a bit on master, and tag a new release.

PR #107, opened 2020-08-06
TenantsArtisanCommand now use TenantAware trait

masterix21:
Refactored the `tenents:artisan` command to use `TenantAware` trait: it reduce the code redundancy.


freekmurze:
Very cool! Thanks!

PR #110, opened 2020-08-07
TenantAware should use the tenant model...

telkins:
I believe that the TenantAware trait for the tenant artisan command should use the configured tenant model.

I haven't run any tests (other than seeing it work when I make the changes locally)....I'm editing it directly in github.  If there are issues, then I'll clone and try to do it "properly".  ;-)


telkins:
Might need to test with a different DB setup....one that uses a custom Tenant model...like I'm doing.  ;-)

masterix21:
@telkins you're right, it's better to use `$this->getTenantModel()`

masterix21:
Thank you

PR #115, opened 2020-08-21
Removed check isCurrent from makeCurrent

masterix21:
It resolves a bug introduced by #111 PR.

Using a job to execute an Artisan command (`Artisan::call(...)`, the Tenant is supplied, but Laravel resets the configuration applied by the tasks. To solve the problem, I removed the lines from 17 to 20 from `Tenant` model.

### DIFF
--- a/src/Models/Tenant.php
+++ b/src/Models/Tenant.php
@@ -14,10 +14,6 @@ class Tenant extends Model
 
     public function makeCurrent(): self
     {
-        if ($this->isCurrent()) {
-            return $this;
-        }
-
         static::forgetCurrent();
 
         $this


masterix21:
I found the problem, and was an arrow function in my code: I solved changing it to a classic closure function. 
Sorry. I will restore the 1.6.4 version.

telkins:
Hey @masterix21 ....thx for the additional input/investigation.  I had hoped to get time over the weekend to look into it some more....but didn't.  :-(

That's fantastic that you found the problem.  Great work...!  🤓 

Anyway, I did sit down this morning and try to find a simple way to show what I believe to **still** be a problem....although not related, perhaps, to what we've been looking at.  Please refer to #117 for a bit more information.

As I stated there (#117 ), I'm hoping someone can help see whether there's actually a problem or not and, if so, where it might be and what the solution might be.

Thx again...!  💪

PR #117, opened 2020-08-24
Simple test to show (potential) issues with package.

telkins:
**PLEASE NOTE THAT THIS PR IS DESIGNED TO INTRODUCE A FAILING TEST**

I've had several problems running unit tests while using this package.  With a previous PRs, I had things working...with some few-but-important exceptions.  @masterix21 recently removed part of some code I had added and now my tests all fail.

I have tried to find a solution (fix the package code or fix my code...or both), but haven't been successful.

So, with some inspiration from @adamwathan (https://twitter.com/adamwathan/status/1297517986576838656) and a little time out of my day today, I set out to try to demonstrate some of my issues by way of a failing test or two....ideally with a solution as well.

I had planned to do something similar to what's found in several existing tests:
- create two tenants
- migrate each tenant database
- create a user for each tenant
- attempt to access each tenant

...all within one test.

I struggled to even get that far.  I found that I could simplify it further.  I took one of the migration tests and added an assertion checking to see if the `users` table could be found.  Turns out it fails.  I'm not sure why or how, but I believe it has something to do with the "built-in" transaction gymnastics that are used to run "clean" tests each time.  I believe switching database connections leads to losing the necessary database transaction context to move forward.

I also looked (briefly) for a solution.  I couldn't find one...yet.  But, I figured it would be worthwhile to submit now and allow for wiser folks to weigh in.

I'm not sure where the problems are with this issue...could be the package, could be my own code or approach to testing, or it could be Laravel's way of working with tests isn't quite compatible with these kinds of scenarios.  Again, I'm not sure.

Anyway, thanks for your patience....looking forward to working through this together.

### DIFF
--- a/tests/Feature/Commands/TenantsArtisanCommandTest.php
+++ b/tests/Feature/Commands/TenantsArtisanCommandTest.php
@@ -51,6 +51,7 @@ class TenantsArtisanCommandTest extends TestCase
 
         $this
             ->assertTenantDatabaseDoesNotHaveTable($this->tenant, 'migrations')
+            ->assertTenantDatabaseHasTable($this->anotherTenant, 'users')
             ->assertTenantDatabaseHasTable($this->anotherTenant, 'migrations');
     }
 


telkins:
I should also note that if/when the migration issue is addressed, I will still look to do the following:
- create two tenants
- migrate each tenant database
- create a user for each tenant
- attempt to access each tenant

...all within one test.

masterix21:
@telkins today I've restored the code removed the last Friday, after a long session of tests.

About the transaction, can you check here #74 to verify if it is related?

telkins:
> @telkins today I've restored the code removed the last Friday, after a long session of tests.

I saw that.  Thx.  I think multitenancy is fantastic, but it does make things tricky when there are problems.  😮 

> About the transaction, can you check here #74 to verify if it is related?

Checking...

telkins:
> > About the transaction, can you check here #74 to verify if it is related?
> 
> Checking...

@masterix21 I have only read through part of the conversation and, yes, IMHO, it seems that this is related.

I got to the part where you suggest he switch to `RefreshDatabase`, which seems to make (some) things a little better for him...but he still gets issues that sound similar to what I've experienced.

I haven't gotten far enough to know if he's implemented another piece of the puzzle: https://docs.spatie.be/laravel-multitenancy/v1/advanced-usage/executing-code-for-tenants-and-landlords/#testing-with-databasetransactions-for-tenant

Alright...caught up, but my brain's a bit mushy.  I've spent more time than I'd care to admit looking into this in my current project.  😝 

So...what I think is happening is related to the `RefreshDatabase` functionality, when transactions are started/stopped, and when database connections are made/purged.

When a (normal) test is started, a transaction is also started.  Whenever a new database connection is made, I believe it is unaware of whatever's taken place within the aforementioned transaction context.

Here's what's executed on every **new** tenant switch (based on the latest release where `makeCurrent()` returns the current tenant if they're the same):
```php
        $tenantConnectionName = $this->tenantDatabaseConnectionName();

        if ($tenantConnectionName === $this->landlordDatabaseConnectionName()) {
            throw InvalidConfiguration::tenantConnectionIsEmptyOrEqualsToLandlordConnection();
        }

        if (is_null(config("database.connections.{$tenantConnectionName}"))) {
            throw InvalidConfiguration::tenantConnectionDoesNotExist($tenantConnectionName);
        }

        config([
            "database.connections.{$tenantConnectionName}.database" => $databaseName,
        ]);

        DB::purge($tenantConnectionName);
```

I'm a bit out of my depth here, but I believe that `DB::purge()` wipes out the current database connection information and then a new connection is made (when needed) for the *new* tenant.

My **guess** is that the new connection doesn't have the same transaction context...which makes sense and shouldn't cause problems because it's a different database/schema.

...**but**...when switching **back** to the first tenant, then the aforementioned code with `DB::purge()` is called.  Now, we're back to the earlier connection....**but without** the earlier transaction context.  So, it has not awareness of whatever took place before.

I'm sure I have some details wrong and it's also possible I'm barking up the wrong tree entirely.  But, I think it has something to do with these multiple connections **and** multiple transaction contexts that are unaware (by design, of course) of each other.

Again, I hope my PR (#117 ) helps to demonstrate this in a simple, straightforward manner.  But, if it doesn't then I'd still love to see a single unit test that can switch from one tenant to another, creating resources in each, then switch back and access said resources.

masterix21:
What do you think [about it](https://github.com/spatie/laravel-multitenancy/issues/74#issuecomment-664685490)?

telkins:
> What do you think [about it](https://github.com/spatie/laravel-multitenancy/issues/74#issuecomment-664685490)?

I don't think it's enough....but I have an idea that I will try out....  🤓

masterix21:
Ok, I've found the problem. Tenant migration doesn't execute any migration (might be a choice of @freekmurze).

```php
/** @test */
    public function it_can_migrate_a_specific_tenant()
    {
        $tenantMigrationsPath = realpath(__DIR__ . '/../../database/migrations');
        $tenantMigrationsPath = str_replace('\\', '/', $tenantMigrationsPath);

        $this->artisan('tenants:artisan "migrate --database=tenant --path='. $tenantMigrationsPath .' --realpath" --tenant=' . $this->anotherTenant->id . '"')->assertExitCode(0);

        $this
            ->assertTenantDatabaseDoesNotHaveTable($this->tenant, 'migrations')
            ->assertTenantDatabaseHasTable($this->anotherTenant, 'users')
            ->assertTenantDatabaseHasTable($this->anotherTenant, 'migrations');
    }
```

Now it works :)

telkins:
> Ok, I've found the problem. Tenant migration doesn't execute any migration (might be a choice of @freekmurze).
> 
> ```
> /** @test */
>     public function it_can_migrate_a_specific_tenant()
>     {
>         $tenantMigrationsPath = realpath(__DIR__ . '/../../database/migrations');
>         $tenantMigrationsPath = str_replace('\\', '/', $tenantMigrationsPath);
> 
>         $this->artisan('tenants:artisan "migrate --database=tenant --path='. $tenantMigrationsPath .' --realpath" --tenant=' . $this->anotherTenant->id . '"')->assertExitCode(0);
> 
>         $this
>             ->assertTenantDatabaseDoesNotHaveTable($this->tenant, 'migrations')
>             ->assertTenantDatabaseHasTable($this->anotherTenant, 'users')
>             ->assertTenantDatabaseHasTable($this->anotherTenant, 'migrations');
>     }
> ```
> 
> Now it works :)

OK.  I'll work on the next test, then, after I've added your change/fix.  Thx...!  👍

telkins:
> Now it works :)

@masterix21 I've now decided to try to approach this in a different way so that you can see exactly what I'm dealing with.

I'm not sure that this conversation should continue here or in an issue or what.  Please let me know.

Anyway, I've created a repository that demonstrates what I'm doing in my current project....and will hopefully flush out an error either in my code/approach or in the package.  **To be clear, each time a new tenant is created, I'm provisioning a new database for it and migrating it...within a job.**

Here's the repo: https://github.com/telkins/laravel-multitenancy-test

If you pull it and spin it up, then when you run the test the first time, it will pass.  After that, it doesn't.

I have some ideas about where to go from here, but I'm not sure what the right solution is.  Also note in the test class that the `refreshTestDatabase()` method is overridden.  This **seems** necessary....but I don't know what you guys might recommend.

From there....not sure where to go.  I don't want to `migrate:fresh` when migrating new tenant databases because if there's ever an error where the tenants get mixed up somehow, then the tenant database won't get wiped out, at least.

I'm starting to wonder, though, if the `refreshTestDatabase()` method should try to find any test tenants and perform a tenant migration...?

All of this is building up to my current issues with trying to test certain scenarios.  I get weird failures and I'm not 100% sure what or where the problem is.

Alright...I'm done for now.  I'll wait for some feedback.

masterix21:
@telkins, it's easy to understand why it happens. If you reset the landlord database, there is no way to understand what are the tenant databases.

You should use it:
```php
protected function refreshTestDatabase()
    {
        if (! RefreshDatabaseState::$migrated) {
            Tenant::all()->eachCurrent(function () {
                $this->artisan('migrate:fresh --database=tenant');
            }); // <- I've added it: please note that if you use the arrow function syntax, it won't works.

            $this->artisan('migrate:fresh', [
                '--drop-views' => $this->shouldDropViews(),
                '--drop-types' => $this->shouldDropTypes(),
                '--path' => 'database/migrations/landlord',
                '--database' => 'landlord',
            ]);

            $this->app[Kernel::class]->setArtisan(null);

            RefreshDatabaseState::$migrated = true;
        }

        $this->beginDatabaseTransaction();
    }
```

telkins:
> @telkins, it's easy to understand why it happens. If you reset the landlord database, there is no way to understand what are the tenant databases.
> 
> You should use it:
> 
> ```
> protected function refreshTestDatabase()
>     {
>         if (! RefreshDatabaseState::$migrated) {
>             Tenant::all()->eachCurrent(function () {
>                 $this->artisan('migrate:fresh --database=tenant');
>             }); // <- I've added it: please note that if you use the arrow function syntax, it won't works.
> 
>             $this->artisan('migrate:fresh', [
>                 '--drop-views' => $this->shouldDropViews(),
>                 '--drop-types' => $this->shouldDropTypes(),
>                 '--path' => 'database/migrations/landlord',
>                 '--database' => 'landlord',
>             ]);
> 
>             $this->app[Kernel::class]->setArtisan(null);
> 
>             RefreshDatabaseState::$migrated = true;
>         }
> 
>         $this->beginDatabaseTransaction();
>     }
> ```

Well, it's OK, but it doesn't explain why the transactions aren't rolled back on the tenant databases.  I think even if I change the code to match what you've done, I'll still encounter the problem further along.

As I understand it, `refreshTestDatabase()` starts the database transaction for all connections listed.  So, when the following is used:
```php
    protected function connectionsToTransact()
    {
        return [
            $this->landlordDatabaseConnectionName(),
            $this->tenantDatabaseConnectionName(),
        ];
    }
```
...then at the **end** of each test, all of the transactions are rolled back for each database connection.

Among other things, one of the problems is that only one tenant database will have its transactions rolled back...at most.

So...I'll add the bit that you've suggested, but then I'll update the test to show the next area/issue that comes along.  🤓

telkins:
@masterix21 I guess the rolling back of transactions doesn't matter...or it happens by default...?  I'm confused about it....but the suggested change seems to be working for the various scenarios that I've struggled with.

So....**awesome**...!!!  🥳 

If I run into any further problems then I'll be back here to report on it and I'll update the repo accordingly.

In the meantime, thanks again for your help/patience.  👍

masterix21:
Rolling-back can't work because the only actor who knows the database is the landlord: but if you refresh it, who knows the database names?

PR #122, opened 2020-08-26
Fix for broken doc links

Braunson:
This PR includes updates to links in the docs that reference itself, it seems the new version of the docs design breaks links within itself.

Perhaps updating the reference in whatever generates the docs on the frontend to include `/docs/` for any internal reference, or just use this PR.

For example:
1. Go to https://spatie.be/docs/laravel-multitenancy/v1/installation/using-multiple-databases
2. Click the text `the base installation steps` link which leads to `https://spatie.be/laravel-multitenancy/v1/installation/base-installation`
3. 404 page.


masterix21:
Thank you

PR #125, opened 2020-09-01
added strict type checking and fixed missing return types and other php fixes

bhushan:
All tests are passing.. 

in many places return types where missing.. added those

enabled strict type checking..


freekmurze:
Thanks, but we are not going to lerge this. We don't use strict type checking in our other packages, and would like to keep consistent.

PR #127, opened 2020-09-01
Subdomain support fixes #126

### LINKED ISSUE #126: Subdomain support
lets suppose I have saas called.. `saas.com` so what i want is when user registers on `saas.com` 
he/she will get access to there tenant from `something.saas.com`

but if they provide there domain with us then it should also work with `theredomain.com`

I hope you are getting my point.

I am working on currently.. Will send PR.. Lets see how far i go..

bhushan:
I have added subdomain finder logic

but i dont know how to hook into the application

Need Help..

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 
 $finder = Symfony\Component\Finder\Finder::create()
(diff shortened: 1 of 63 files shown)


freekmurze:
We don't use strict type checking

PR #128, opened 2020-09-01
Subdomain support without strict type

bhushan:
I have added subdomain finder logic

added missing return types in few places

now i need help in hooking both `DomainTenantFinder` and `SubdomainTenantFinder` in this package @freekmurze 

Let me know things i can improve logic or fix anything.. 

my current idea is .. tenant finder should be like `actions` 

it should first check `subdomain` and if we are not on `landlord domain url` then it should check for `domain`.

```php
  'tenant_finders' => [
      Spatie\Multitenancy\TenantFinder\SubdomainTenantFinder::class,
      Spatie\Multitenancy\TenantFinder\DomainTenantFinder::class
],
```

Let me know what you feel about this idea.

I have used NotFoundHttpException is few places.. in SubdomainTenantFinder

I could not figure out where should it go. I want your opinion on that.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 
 $finder = Symfony\Component\Finder\Finder::create()
     ->notPath('bootstrap/*')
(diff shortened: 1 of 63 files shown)


freekmurze:
Thanks for your work on this, but we're not going to merge it in.

We'd like to keep this package as simple as it can be. We think the current way of handling tenant finders offers enough flexibility for most users.

bhushan:
@freekmurze 

that's Ok.. could you give me idea .. how i can achieve it ?

freekmurze:
I would create a single tenant finder and add all necessary logic in there.

bhushan:
I thing you should close issue as well ..

PR #129, opened 2020-09-01
updated github action versions

bhushan:
While checking package code saw that github actions where old.. 

updated based on https://github.com/stefanzweifel/git-auto-commit-action


freekmurze:
Thanks!

PR #142, opened 2020-09-27
Update MakeQueueTenantAwareAction.php

3assy2018:
Adding Use Line Of Broadcast Event To Detect Jobs From Queueable


masterix21:
Thanks @3assy2018.

PR #144, opened 2020-10-02
Removed @test annotations from the exception class

claudsonm:
Hi there!

I was making a lightweight version of this package for a project and spotted this. It looks like some `@test` annotations ended up in the source code.


masterix21:
Thanks @claudsonm

PR #156, opened 2020-10-28
Update composer.json

merdanbh:
Can you add php 7.3 version support?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": "^7.4",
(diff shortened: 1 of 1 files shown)


masterix21:
I'm sorry, but the package uses PHP 7.4 features like the arrow functions. I don't know what @freekmurze thinks, but if he agrees, you could create a more complex PR with all code parts changed to support PHP 7.3, not composer.json only.

Thanks

freekmurze:
We don't use PHP 7.3 for our projects so we're not going to invest time in that.

PR #158, opened 2020-10-29
Fix for duplicate test

bryanjamesmiller:
I noticed this test is exactly the same as the one before it, and it appears it is just missing creating a tenant with a different domain than the one in the request to show, as the name of the test says, that `it_will_return_null_if_no_tenant_can_be_found_for_the_current_domain`


masterix21:
Thank you

PR #163, opened 2020-11-10
Update using-multiple-databases.md

ivanmagdic:
Document that tenant migrations can also be in specific directory (ex. database/migrations/tenant).


masterix21:
Thank you @ivanmagdic

PR #171, opened 2020-11-24
NeedsTenant should able to return or redirect

gazben:
The `NeedsTenant` middleware `handleInvalidRequest` should not be restricted to a void return value. It should be able to produce a redirect or a view to display an error message.


masterix21:
It's a good idea: I will merge it in the next release.

Thank you for the PR.

masterix21:
Merged. Thanks.

PR #182, opened 2020-12-29
Switch session configuration task

undjike:
Hello, I am submitting this pull request to make it possible to auto-configure sessions when the current tenant changes. By following the example implementation of the other tasks, I was able to produce this modest result. Take a look and let me know if there is any error.

### DIFF
--- a/docs/using-tasks-to-prepare-the-environment/session-config.md
+++ b/docs/using-tasks-to-prepare-the-environment/session-config.md
@@ -1,0 +1,39 @@
+---
+title: Session config
+weight: 4
+---
+
+You might want to manage sessions per tenant. In this case, you can use `Spatie\Multitenancy\Tasks\SwitchSessionConfigTask`.
+
+To use this task, as for others, you should add it to the `switch_tenant_tasks` key in the `multitenancy` config file.
+
+```php
+// in config/multitenancy.php
+
+'switch_tenant_tasks' => [
+    \Spatie\Multitenancy\Tasks\SwitchSessionConfigTask::class,
+    // other tasks
+],
+```
+
+When working with database session driver, you should be aware of the following.
+
+1. The Landlord environment will use the default session driver. So, if that default driver is set to `database`, you should copy the session migration to the `database/migrations/landlord` folder and migrate it. With this done, when no tenant is set, Laravel will use the Landlord environment to manage sessions.
+
+
+2. You probably may not want that complexity. You can set the session default driver to `file` so that no database table will be needed for sessions when no tenant is set. Then, you can add a key in your `config/session.php` file.
+
+```php
+[
+    // other config
+    'tenant_driver' => 'database',
+],
+```
+
+After this has been done, whenever a tenant is set, the database connection defined for tenants will be used for sessions too.
+
+### Notes
+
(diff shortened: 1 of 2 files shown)


masterix21:
Hi @undjike, thanks for your work, but I think it introduces an unneeded complexity. I prefer to remand the final decision to @freekmurze.

docs/using-tasks-to-prepare-the-environment/session-config.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Session config
+weight: 4
```
undjike, line 3: Here, I didn't know what the weight should be. So, it's just a speculative value.

src/Tasks/SwitchSessionConfigTask.php
```diff
@@ -0,0 +1,46 @@
...
+
+    private $originalDriver;
+
+    public function __construct()
+    {
+        $this->originalDriver = config('session.driver');
+        $this->originalConnection = config('session.connection');
+    }
+
+    public function makeCurrent($tenant): void
```
undjike, line 17: Here I did not set the type of the parameter because I assumed that a custom tenant model could be used by the developer using the package.

undjike:
> Hi @undjike, thanks for your work, but I think it introduces an unneeded complexity. I prefer to remand the final decision to @freekmurze.

Okay, no worries.

freekmurze:
I agree with @masterix21 that this adds too much complexity in this page.

The core philosophy of our package is to keep things as simple as possible, and we don't want to do any specific things like switching sessions config.

If you put your code in a separate package, we'll be happy to link to it in our docs 👍

src/Tasks/SwitchSessionConfigTask.php
gazben: This causes problems with larastan, and other static analyzers because the original declaration does not match this line.

src/Tasks/SwitchSessionConfigTask.php
```diff
@@ -0,0 +1,46 @@
...
+class SwitchSessionConfigTask implements SwitchTenantTask
+{
+    private $originalConnection;
+
+    private $originalDriver;
+
+    public function __construct()
+    {
+        $this->originalDriver = config('session.driver');
+        $this->originalConnection = config('session.connection');
```
gazben, line 14: I think this should look like this: `$this->originalConnection = config('session.connection') ?? config('database.default');
`

Because the `session.connection` can be null by default

gazben:
@freekmurze The code itself is pretty small. Can it be included just to the docs? Like this is included: `https://spatie.be/docs/laravel-multitenancy/v1/using-tasks-to-prepare-the-environment/prefixing-cache`

freekmurze:
Maybe I'm too strict about this, but I don't want to maintain it in any way. I'll happily link to it if you display it in another repo / blogpost

src/Tasks/SwitchSessionConfigTask.php
undjike: Sorry, I don't see your point... Any problem if it is `null`?

gazben: The `setSessionDatabaseConnection` function only accepts strings. It will fail on `null`

PR #188, opened 2021-01-13
Add support for executing delayed callbacks in Tenant context

drasko95:
I have added a commit for allowing delayed callbacks to be executed inside of the Tenant context. It is useful in case when callbacks are defined in the Tenant context but their execution takes place outside of it. One notable example of this use case is the Laravel scheduler.

For example, the following code inside the `schedule()` function won't work correctly, because even though the callback is defined in the tenant context, it is simply being added to the queue and is executed after the `schedule()` function terminates:

```php
Tenant::all()->eachCurrent(function(Tenant $tenant) use ($schedule){
    $schedule->run(fn() => /*do something*/)->Daily();
})
```

With this proposal, the desired behavior can be achieved like this

```php
Tenant::all()->eachCurrent(function(Tenant $tenant) use ($schedule){
    $schedule->run($tenant->callback(fn() => /*do something*/))->Daily();
})
```

This will make sure the code inside the callback is executed inside the correct Tenant's context.

### DIFF
--- a/tests/Feature/Models/TenantTest.php
+++ b/tests/Feature/Models/TenantTest.php
@@ -157,4 +157,26 @@ class TenantTest extends TestCase
 
         $this->assertEquals($response, $this->tenant->id);
     }
+
+    /** @test */
+    public function it_will_execute_a_delayed_callback_in_tenant_context()
+    {
+        Tenant::forgetCurrent();
+
+        $this->assertNull(Tenant::current());
+
+        $callback = $this->tenant->callback(function (Tenant $tenant) {
+            $this->assertEquals($tenant->id, Tenant::current()->id);
+
+            return $tenant->id;
+        });
+
+        $this->assertNull(Tenant::current());
+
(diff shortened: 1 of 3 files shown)


masterix21:
It's good to use a readable callback. It's ok for me, but I prefer to wait for the @freekmurze opinion before merging it.

In the meanwhile, if you can, please write the documentation. Thanks.

tests/Feature/Models/TenantTest.php
```diff
@@ -157,4 +157,26 @@ public function it_will_execute_a_callable_and_then_restore_the_previous_state()
 
         $this->assertEquals($response, $this->tenant->id);
     }
+
+    /** @test */
+    public function it_will_execute_a_delayed_callback_in_tenant_context()
```
freekmurze, line 162: Very nice test 👍

freekmurze (approved):

masterix21:
Merged. Thanks

PR #223, opened 2021-04-07
Laravel Octane support

masterix21:
- [x] Find the right Tenant using the Laravel Octane events
- [x] Verify switch database feature
- [x] Verify if the configurations are "tenant isolated" with crossed requests
- [x] Check queues

@freekmurze, I'm working to support Laravel Octane. What's your idea or your suggestions?

Landlord response without Octane:
image

Landlord response with Octane:
image

Tenant response without Octane:
image

Tenant response with Octane:
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "mockery/mockery": "^1.4",
         "orchestra/testbench": "^6.0",
         "phpunit/phpunit": "^9.4",
-        "spatie/valuestore": "^1.2"
+        "spatie/valuestore": "^1.2",
+        "laravel/octane": "^0.1.1"
     },
     "autoload": {
         "psr-4": {
--- a/config/multitenancy.php
+++ b/config/multitenancy.php
@@ -12,11 +12,6 @@ use Spatie\Multitenancy\Models\Tenant;
 
 return [
     /*
-     * If your application uses Laravel Octane, set it to `true`.
-     */
-    'octane' => false,
-
-    /*
      * This class is responsible for determining which tenant should be current
      * for the given request.
      *
--- a/config/multitenancy.php
+++ b/config/multitenancy.php
@@ -12,6 +12,11 @@ use Spatie\Multitenancy\Models\Tenant;
 
 return [
     /*
+     * If your application uses Laravel Octane, set it to `true`.
+     */
+    'octane' => false,
+
+    /*
      * This class is responsible for determining which tenant should be current
      * for the given request.
      *
--- a/src/Multitenancy.php
+++ b/src/Multitenancy.php
@@ -1,0 +1,94 @@
+<?php
+
+namespace Spatie\Multitenancy;
+
+use Illuminate\Contracts\Foundation\Application;
+use Illuminate\Support\Facades\Event;
+use Laravel\Octane\Events\RequestReceived as OctaneRequestReceived;
+use Laravel\Octane\Events\RequestTerminated as OctaneRequestTerminated;
+use Laravel\Octane\Octane;
+use Spatie\Multitenancy\Actions\MakeQueueTenantAwareAction;
+use Spatie\Multitenancy\Concerns\UsesMultitenancyConfig;
+use Spatie\Multitenancy\Models\Concerns\UsesTenantModel;
+use Spatie\Multitenancy\Models\Tenant;
+use Spatie\Multitenancy\Tasks\TasksCollection;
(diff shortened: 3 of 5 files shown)


config/multitenancy.php
```diff
@@ -11,6 +11,11 @@
 use Spatie\Multitenancy\Models\Tenant;
 
 return [
+    /*
+     * If your application uses Laravel Octane, set it to `true`.
+     */
+    'octane' => false,
```
freekmurze, line 17: It would be cool we could just detect Octane. @themsaid is there something in Laravel/Octane that allows easy detection?

freekmurze (approved):
Looks good! Amazingly speedy results!

config/multitenancy.php
themsaid: You can do `isset($_SERVER['LARAVEL_OCTANE'])` to check if it's running under Octane.

masterix21: Thank you

masterix21:
UPDATE NOTICE

Currently seems that `DB::purge('tenant');` and `DB::reconnect('tenant');` aren't enough to switch the tenant database connection. 

The only way that I've found to solve the database connection issue is:
```php
app('db')->extend('tenant', function ($config, $name) {
     $config['database'] = 'laravel_mt_tenant1';

     return app('db.factory')->make($config, $name);
});
```

@themsaid, do you have a suggestion?

themsaid:
@masterix21  why not enough?

masterix21:
@themsaid see the code below.

Original `SwitchTenantDatabaseTask` created by @freekmurze, Octane doesn't switch the connection.

image
image


Instead, using the code like so, works:

image
image

I know that what works is a solution to a problem, but I'd like to understand if there are other ways to resolve the issue.

composer.json
```diff
@@ -25,7 +25,8 @@
         "mockery/mockery": "^1.4",
         "orchestra/testbench": "^6.0",
         "phpunit/phpunit": "^9.4",
-        "spatie/valuestore": "^1.2"
+        "spatie/valuestore": "^1.2",
+        "laravel/octane": "^0.1.1"
```
ju5t, line 29: I guess this one should go when the PR gets merged?

masterix21: Yes, but it's only for dev purpose, because isn't required by the package.

masterix21:
@freekmurze, if for you is ok, I think this could be the first release Octane ready. I made changes to the Service Provider, which are required to support Octane: if you have a better approach I'm here.

Thank you.

src/Multitenancy.php
```diff
@@ -0,0 +1,90 @@
...
+use Spatie\Multitenancy\Models\Tenant;
+use Spatie\Multitenancy\Tasks\TasksCollection;
+use Spatie\Multitenancy\TenantFinder\TenantFinder;
+
+class Multitenancy
+{
+    use UsesTenantModel,
+        UsesMultitenancyConfig;
+
+    public function __construct(public Application $app)
```
freekmurze, line 18: 👍

freekmurze (approved):

freekmurze:
Looks good to me, thanks for you work on this! 👍

bcalik:
@masterix21 Can we utilize multiple database connections for each tenant instead of purging the existing connection with each request using Octane? This could significantly boost the performance I guess.

So firstly [tenantDatabaseConnectionName()](https://github.com/spatie/laravel-multitenancy/blob/main/src/Concerns/UsesMultitenancyConfig.php#L10) should return different connection name depending on the request. Then I'm not sure but there might be some breaking changes, but why not make it for v4?

masterix21:
What happens if you have many tenants? Many (in-memory) opened connections? I wonder if this is the right way.

bcalik:
> What happens if you have many tenants? Many (in-memory) opened connections? I wonder if this is the right way.

Yes, but I think it’s not a huge deal breaker. We would need to manage the maximum number of connections by setting timeouts for inactive connections.

By the way, we already need to handle hundreds of open connections using FPM per request for a high-traffic web app. Since Octane will only open a single connection for a single tenant, you could easily handle a very profitable business.

I also think that by opening and closing connections, we lose the ability to handle a significant number of requests per second. We waste more CPU, and CPU time is expensive, whereas memory is cheap. So, I prefer keeping more connections open.

patie:
Any updates here? Im curious if this packagebis fully octane(swoole) compatible and also how db connections are managed. Thank you!

PR #256, opened 2021-06-10
Update link to Tom's Blog Post

Illizian:
It seems Tom may have updated the routes on their blog without redirects :grimacing:, so here's the new URL.


masterix21:
Thanks

PR #274, opened 2021-08-03
Add tenant scope to improve the single-database mode

masterix21:
🚧 Work in progress: code not tested 🚧

The package has a lot of features for the multi-database mode but is poor with the single mode. 

The feature improves our support to the single-mode with a `TenantScope` and a trait `BelongsToTenant`. 

For example, imagine a model `Setting` that belongs to a Tenant.

```php
class Setting extends Model
{
    use BelongsToTenant;
}
``` 

Now, calling the class from your code, the tenant foreign key will always be present: and the same will be when you create a new `Setting`.

Finally, if for any coding reason you need to remove the global scope, you can use the Laravel feature:
```php
Setting::withoutGlobalScope(TenantScope::class)->get()
```

@freekmurze what do you think about it? I will create the tests and the documentation if is ok for you.

### DIFF
--- a/src/Models/Concerns/BelongsToTenant.php
+++ b/src/Models/Concerns/BelongsToTenant.php
@@ -4,7 +4,10 @@ namespace Spatie\Multitenancy\Models\Concerns;
 
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Spatie\Multitenancy\Exceptions\NoCurrentTenant;
 use Spatie\Multitenancy\Models\Concerns\Scopes\TenantScope;
+use Spatie\Multitenancy\Models\Tenant;
+use Spatie\Multitenancy\Multitenancy;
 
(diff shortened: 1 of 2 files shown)


freekmurze:
@masterix21 Is there are real world use-case or did some ask for this feature? If nobody is asking or is needing this, I'd like to keep the package as is for now.

masterix21:
The idea starts from here:

https://github.com/spatie/laravel-multitenancy/discussions/269

So, now I’m working on a small project with a single database, and I think that the PR is for the single database what for the multi-database is the trait UsesTenantConnection.

freekmurze:
@masterix21 Thanks! Do continue working on this, add tests + docs and I'll review it!

masterix21:
Finally, I think is better to marry your idea to merge all these pieces of information in the docs instead of additional package complexity. I will work on the docs update as soon as possible. Sorry for it, but these days I'm a bit busy.

PR #288, opened 2021-09-10
Describe option that does *not* default to all tenants...

telkins:
I wanted to update the docs to reflect the difference between requiring a specified tenant versus allowing for **all** tenants to be specified by default.

I'm not sure if this is correct/possible/advisable/whatever, but I ran into the problem when creating a very destructive command that I want to use carefully.  I don't want to accidentally run the command against all tenants by simply forgetting to specify a particular tenant.  I changed the signature to use `{--tenant=}` instead of `{--tenant=*}` and it appears to work as expected.  That is, it **requires** a tenant to be specified if none is provided.

Hopefully the use case makes sense and hopefully the use of `{--tenant=}` instead of `{--tenant=*}` is OK.  Please let me know if I'm wrong.

I also changed the docs quickly, so it may not be worded well or thorough enough.  I figured more effort could be put into it once I got the green light from Spatie that this was a reasonable change/use. 🤓

I look forward to your feedback...

### DIFF
--- a/docs/advanced-usage/making-artisan-commands-tenant-aware.md
+++ b/docs/advanced-usage/making-artisan-commands-tenant-aware.md
@@ -3,7 +3,9 @@ title: Making Artisan command tenant aware
 weight: 3
 ---
 
-Commands can be made tenant aware by applying the `TenantAware` trait. When using the trait it is required to append `{--tenant=*}` to the command signature.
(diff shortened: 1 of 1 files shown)


masterix21:
It's never too late to improve the docs. Thanks!

telkins:
> It's never too late to improve the docs. Thanks!

Thx, @masterix21 .  I just want to make sure that this is a reasonable way to use the package, that it makes sense, and that it's a fair/reasonable use case.  It **seems** like the answer should be YES to each point, but I was hoping to hear that from someone "official".  🤓

PR #298, opened 2021-10-15
Handle `JobRetryRequested` queue event and fix #259

### LINKED ISSUE #259: Queue retry fails for tenant aware jobs
Hello, 

Forgive me if I'm wrong, but I think that there is an issue when trying to retry failed tenant aware jobs.

- configuration `queues_are_tenant_aware_by_default` set to `true`
- Not trying to add the jobs into the tenant databases (most of issues are related to this)

1. Typical failed job retry

`php artisan queue:retry ID` fails because it can not unserialize the payload command at [\Illuminate\Queue\Console\RetryCommand::refreshRetryUntil](https://github.com/laravel/framework/blob/ea695bf5974c90347264efdb5f724562846cbe62/src/Illuminate/Queue/Console/RetryCommand.php#L156-L183). Switch tenant database never called...

2. Retry job via tenant aware command - **NO** tenant specified

`php artisan tenants:artisan "queue:retry ID"` partially works but it's executed for every tenant which means that command will succeed only once and will fail for all other tenants. 

3. Retry job via tenant aware command - tenant specified

`php artisan tenants:artisan "queue:retry ID" --tenant=TENANT_ID` works as it should but this means that every time we should check for the `TENANT_ID` at the payload which is nasty ;). When you have a lot jobs failed it's even worse...

Possible solution could be if we have tenant aware queue retry command that could extend `\Illuminate\Queue\Console\RetryCommand` and wrap `retryJob` parents method with set/forget tenant calls if it's a tenant aware job of course.

What do you think about this solution? Any better ideas?

aginev:
* listenForJobsRetryRequested method has been added and executed
* Because JobProcessing and JobRetryRequested has no shared interface a new getEventPayload method has been presented to be able to extract the payload from incompatible classes
* findTenant modified to be able to get the payload correctly

### DIFF
--- a/src/Actions/MakeQueueTenantAwareAction.php
+++ b/src/Actions/MakeQueueTenantAwareAction.php
@@ -3,6 +3,7 @@
 namespace Spatie\Multitenancy\Actions;
 
 use Illuminate\Queue\Events\JobProcessing;
+use Illuminate\Queue\Events\JobRetryRequested;
 use Illuminate\Support\Arr;
 use Spatie\Multitenancy\Exceptions\CurrentTenantCouldNotBeDeterminedInTenantAwareJob;
 use Spatie\Multitenancy\Jobs\NotTenantAware;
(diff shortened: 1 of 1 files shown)


masterix21:
@aginev, could you add the test cases, please?

PR #308, opened 2021-11-26
Add route cache path switcher

AlexVanderbist:
- adds `SwitchRouteCachePathTask` + test
- adds all included tenant switchers to config file (commented out)
- (adds new php cs fixer cache path to gitignore


freekmurze:
Flawless victory!

masterix21:
@AlexVanderbist great PR.

masterix21:
It partially works because fails with a code like so:

```php
//
// Root path for landlord
Route::domain('landlord.test')
    ->get('/', fn () => view('welcome-landlord'));

// 
// Root path for tenant(s)
Route::middleware('tenant')
    ->domain(Tenant::current()?->host ?? 'tenant.test')
    ->get('/', fn () => view('welcome-tenant'));
```

The above scenario seems to deny the routes caching. 

Do you have an idea to avoid the use of a `/landlord/` path prefix?

masterix21:
> "However, if you're using dynamic routes, for example different routes for different tenants, you'll need to keep a separate route cache for each tenant."

Laravel `route:cache` uses a fresh `$app` object, and if you try to push new routes from a tenant task, it won't work. Do you have found another way?

masterix21:
That’s the solution to the problem.

**1. Create a new task to build the tenant routes**
```php
namespace App\Tenants\Tasks;

use Illuminate\Support\Facades\Route;
use Spatie\Multitenancy\Tasks\SwitchTenantTask;

class SwitchRoutesTask implements SwitchTenantTask
{
    public function makeCurrent(Tenant $tenant): void
    {
        if (app()->routesAreCached()) {
            return;
        }

        Route::middleware(['api', 'tenant'])
            ->domain($tenant->host)
            ->group(base_path('routes/tenants/api.php'));

        Route::middleware(['web', 'tenant'])
            ->domain($tenant->host)
            ->group(base_path('routes/tenants/web.php'));
    }

    public function forgetCurrent(): void
    {
        // I think we should clear the router instance, but currently isn't essential for me.
    }
}

```

**2. Update the `multitenancy.php` configuration like so:**
```php
'switch_tenant_tasks' => [
    // ...
    \Spatie\Multitenancy\Tasks\SwitchRouteCacheTask::class,
    \App\Tenants\Tasks\SwitchRoutesTask::class,
    // ...
],
```

**3. Cache your routes**
```sh
php artisan tenant:artisan route:cache
```

That's all. 

@freekmurze / @AlexVanderbist,  I think it should be helpful in the documentation. Do you think so too?

PR #309, opened 2021-11-29
Reload Router instance when switching route cache on Laravel Octane

AlexVanderbist:


AlexVanderbist:
We can probably do better if we don't forget the current tenant at the end of every Octane request. However, that's out of scope.

PR #312, opened 2021-12-02
[Proposal] Issue templating

masterix21:
Hi dears, 

to improve our repository organization I think is better to introduce the issue templating offered by GitHub.

Any new issue will show a screen like so:
example

It will help the users to use the discussions instead of issues for questions and ideas.

What do you think about it?


freekmurze:
Looks good to me, we also added this to [our skeleton repo](https://github.com/spatie/package-skeleton-laravel) too.

Issue #305, opened 2021-11-09
multitenancy with Jetstream / Fortify - database selection / session issue


CodeAdminDe:
Hello together,

## Initial situation

I want to make a Laravel application multitenant capable with the 'spatie/laravel-multitenancy' package and use multiple databases (landlord, tenant1, tenant2, ...). The app is a Jetstream installation (installed without --team option) and uses Livewire as a base. 'database' is used as session driver (Jetstream default).

## Steps performed

I followed the steps in the installation guide at https://spatie.be/docs/laravel-multitenancy/v2/installation/base-installation to install the package, setup the config and database connections and added the `UsesTenantConnection/UsesLandlordConnection` traits to my models. 
Since all routes should be tenant aware, I added the package middlewares to the "web" route group.

For testing purposes I've created a tenant incl. database manually ('tenant1.localhost').

## Issue

The sessions are written to / read from the landlord database and not - as I would expect - to / from the tenant databases. 
This can be recreated by calling 'http://tenant1.localhost' as this fails until I perform the migration of the session table for the landlord database. Also sessions of other tenants are written to the landlord database as soon as the migration was done.

![grafik](https://user-images.githubusercontent.com/53627251/140996135-306666c9-427b-4e37-8d62-8ccca40d55f9.png)

Furthermore, I noticed that Livewire also tries to write tenant aware data to the landlord database. (Tested by saving changes at the default jetstream profile information component).

![grafik](https://user-images.githubusercontent.com/53627251/140997290-4e564b2f-82f6-4bde-8f7b-2a2d8004c4bf.png)

How can I get Fortify(?) to store / read the sessions tenant aware in the respective tenant database? (Changing the driver to file would solve the issue supposedly, but this is not an option for other reasons at all).

How do I get Livewire (or globally all components) to use the correct databases as well?

I hope the description of the issue is clear enough and appreciate any " push in the right direction". 
Otherwise, please ask questions and I'll try to answer them as best I can.

Thanks in advance!

masterix21:
Did you try to use `SESSION_CONNECTION=landlord` in your .env file?

CodeAdminDe:
> Did you try to use `SESSION_CONNECTION=landlord` in your .env file?
> 
> I'm closing here becuase isn't a package issue, but feel free to continue the talk here.

Sorry for the long response time. Life got in the way. :/

[...]

PR #314, opened 2021-12-18
Added 'Setting session connection' section

CodeAdminDe:
I'd like to suggest the addition within my pull request to include a note that it may be necessary to set the SESSION_CONNECTION variable correctly when using database based sessions.

The idea came from avoiding further issues like mine. (https://github.com/spatie/laravel-multitenancy/issues/305) ;-)

Thanks in advance!

### DIFF
--- a/docs/installation/using-multiple-databases.md
+++ b/docs/installation/using-multiple-databases.md
@@ -143,6 +143,14 @@ class DatabaseSeeder extends Seeder
 
 All models in your project should either use the `UsesLandlordConnection` or `UsesTenantConnection`, depending on if the underlying table of the models lives in the landlord or tenant database.
 
+### Setting session connection
+
+You should set the environment variable `SESSION_CONNECTION` within your `.env` file to use your configured tenant connection, if using database based sessions for each tenant (e.g. with Laravel Jetstream / Fortify).
+
+```
+SESSION_CONNECTION=tenant
+```
+
 ### Next steps
 
 When using multiple tenants, you probably want to [isolate the cache](/docs/laravel-multitenancy/v2/using-tasks-to-prepare-the-environment/prefixing-cache/) or [use separate filesystems per tenant](/docs/laravel-multitenancy/v2/using-tasks-to-prepare-the-environment/filesystem/), ... These things are performed by task classes that will be executed when making a tenant the current one.


masterix21:
I’m not sure about it. What happens if you like to handle the sessions in a landlord context?

One database reserved for the sessions is still the best way, for me.

CodeAdminDe:
I'd like to isolate each tenants data as much as possible from each other. From my point of view, it is

* is easier to identify sessions for a specific tenant
* avoid conflicts between sessions of other tenants
* tenant / sesitive data is stored only in the tenant-aware database

I'd like to use the landlord connection only for "routing" purpose.

I don't get the point why it would be better to put the sessions into the landlord database within a multi database setup. 
Could you tell me the benefits?

masterix21:
There is no benefit than other solutions, but IMHO documentation should drive the user on best practices. 

Merging the PR, I’m sure we should fight many issues on GitHub, for example with who likes to use the sessions in the landlord context.

I’m really don’t feel like merging this, so I prefer to listen the opinion of @freekmurze about it.

CodeAdminDe:
I've found another issue with the approach to set the `SESSION_CONNECTION=tenant` env value:

Every request that does not have a valid tenant loaded, will result in `SQLSTATE[3D000]: Invalid catalog name: 1046 No database selected ...` because there is no database to store the session.

So i'd like to close the PR without merging it.

To solve this issue, i see two options

1. Using the landlord database for every session (like you mentioned).
2. Pushing the middleware '\Spatie\Multitenancy\Http\Middleware\NeedsTenant::class' before '\Illuminate\Session\Middleware\StartSession::class' within the web middlwareGroups.

Hope it is okay to ask two further questions, to improve understanding and to make the right decisions:

1. Do you see any security relevant issues (like takeing over a session or accidental overlapping session ids) when using the landlord database for all tentant sessions? (especially since for example the user_id can overlap between tenants of different databases).

2. Did I understand that correctly?:

> By using the middleware `\Spatie\Multitenancy\Http\Middleware\EnsureValidTenantSession::class` it is ensured that the sessions are *always* scoped to the logged in user tenant, so that a transfer between tenants - even when stored in only one database - is not possible.

> The scoping eliminates the possibility of user_id overlap and problems due to overlapping session ids are not possible (since the tenant unique user_id is also used for session identification).

If I have understood it correctly, this allays my security concerns and I will switch to the variant recommended by you and store the sessions exclusively in the landlord database.

Thanks in advance for your input and your effort!

masterix21:
Take a look here:
[Protecting against cross tenant abuse by Spatie docs.](https://spatie.be/docs/laravel-multitenancy/v2/installation/base-installation#protecting-against-cross-tenant-abuse)

To improve your database session, you could create a custom session driver that store the tenant_id in your sessions table.
[Adding a custom session driver by Laravel docs.](https://laravel.com/docs/8.x/session#adding-custom-session-drivers)

CodeAdminDe:
Thanks for pointing to the custom session handler idea!

I'm pretty sure that's the best way to keep security high and at the same time not limit the functions. :-)

I've created a custom session handler on top of the original DatabaseSessionHandler to avoid unexpected conflicts between sessions by scoping the read/write methods to the tenant_id (or null if no tenant is available). For reference and to show others (and my future self) a possible solution as well, i have documented an example implementation here: https://gist.github.com/CodeAdminDe/ae81bf9765e25ccfcdae51369088bfd0

PR #315, opened 2021-12-20
Update base-installation.md

fullbl:
I think this could be useful for people using the package with Lumen!


masterix21:
@fullbl, thanks for your time on it. 

I have no idea if the package works with Lumen: are you using it with Lumen?

freekmurze:
We're not going spend time on Lumen, as we feel people shouldn't use that anymore.

PR #326, opened 2022-02-06
docs: change method visibility in example

medvinator:
Cause method visibility shouldn't be overridden.
Original is `protected`

https://github.com/laravel/framework/blob/2246744061de4bfd2d3a42d715e6b760015fa074/src/Illuminate/Database/Eloquent/Model.php#L298-L308


masterix21:
Hi @medvinator, could you dig with tests? Seems fails.

medvinator:
I changed the documentation file. It should not affect the tests in any way 🤷‍♂️

masterix21:
Sorry. Thanks.

PR #341, opened 2022-03-17
docs: Updated custom tenant model

ZigaStrgar:
### Description

Updating documentation related to using a custom tenant model. I've overridden `$table` property on model, as it'd by default point to `custom_tenant` table. 

And updated config reference to match the namespace in the example.


masterix21:
Hi @ZigaStrgar, thanks for your time on it, but I think it isn't necessary.

PR #346, opened 2022-03-25
Add SwitchRootUrlTask

francoism90:
Hi!

I don't think you guys are interested in this merge, however I'm also open to add documentation instead. :)

Just let me know, and thanks again for your packages!


masterix21:
Hi @francoism90, thanks for your time on it, but I think that it introduces a complexity unneeded. 

I really appreciate your work on it, but I think could be better to include these instructions in the documentation: a PR for it is good.

Thanks, again.

PR #349, opened 2022-03-28
Remove link to filesystem doc that doesn't exist

skoontastic:
I wasn't able to find the `using-tasks-to-prepare-the-environment/filesystem.md` page, and wasn't qualified to write it, so I thought we should remove it for now until someone can write up how to handle switching filesystems.


masterix21:
Thanks for your time on it.

PR #351, opened 2022-03-30
Fix cached routes

masterix21:
The PR fixes a bug when the package is running in a console environment (tinker too) or unit tests and introduces the configurable ability to create only one routes cache file instead of a file for each tenant.

### DIFF
--- a/config/multitenancy.php
+++ b/config/multitenancy.php
@@ -69,6 +69,12 @@ return [
      */
     'current_tenant_container_key' => 'currentTenant',
 
+    /**
+     * Set it to `true` if you like to cache the tenant(s) routes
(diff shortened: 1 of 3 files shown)


masterix21:
@freekmurze, take a look here: fixes a bug and adds the ability to use the shared route cache file for all tenants.

The shared route cache is excellent to help the users with landlord routes (for example: `routes/landlord.php`) and tenants routes (for example, `routes/tenants.php`). I had the issue in a project of mine, solved overriding the package task, but I'm sure that will help many other users.

masterix21:
Already fixed with another PR: the issue was in the migration tests.

PR #359, opened 2022-04-11
Update create_landlord_tenants_table.php.stub

francoism90:
A database can be shared between tenants. For instance `account.domain.com` and `domain.com` can share the same database.


masterix21:
Thanks for your time on it, but I prefer to keep the migration as already it is: anyone can modify the migration file easily, project by project.

PR #361, opened 2022-04-24
Docs: fix the link to the video introduction

PovilasKorop:
Not sure if that's the correct link, but the old link leads to "Page not found" on Spatie.be website.
That "Page not found" page, by the way, is also indexed on google by "laravel spatie multi tenancy", so you may want to do something with that page, too.


masterix21:
Thanks for your time on it.

PR #374, opened 2022-06-27
Prefix Cache Task - Cache Store

moisish:
I am using your package with the PrefixCacheTask along with the Unique Jobs feature on laravel and I have realised that the lock on the jobs does not get released.

The underline code of laravel is not using the Cache Manager but it's using the Cache Repository.

```
    protected function ensureUniqueJobLockIsReleased($command)
    {
         ....
        $cache = $this->container->make(Cache::class);

        $cache->lock(
            'laravel_unique_job:'.get_class($command).$uniqueId
        )->forceRelease();
    }
    
```

The Cache Repository holds an old version of the CacheManager which does not have the correct cache prefix and therefore none of the locks are deleted.

### DIFF
--- a/src/Tasks/PrefixCacheTask.php
+++ b/src/Tasks/PrefixCacheTask.php
@@ -41,6 +41,9 @@ class PrefixCacheTask implements SwitchTenantTask
         // the `$store` but doesn't update the `$app['config']`.
         app()->forgetInstance('cache');
 
+        //This is important because the Cache Repository is using an old version of the CacheManager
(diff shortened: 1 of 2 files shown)


masterix21:
Could you add the tests?

moisish:
I have added some tests. Let me know if you it's good now

moisish:
@freekmurze can you check this please

masterix21:
@moisish, please fix the errors in the tests. Thanks

PR #394, opened 2022-09-20
Authentication Based Multi-Tenancy

kassah:
This PR adds the middleware required for authentication based multi-tenancy, as well as documentation on how to use it.

#### Example of it's application utilizing the documentation:
https://github.com/kassah/multitenancy-pr-demo

The README in the repository covers demonstration steps.

#### Commit that makes the changes following documentation 
https://github.com/kassah/multitenancy-pr-demo/commit/84d9a80afd8505b9f83b85838b6cb9e661eca929

### Dependency Change
Increased the minimum version of Mockery, so that tests with mocks utilizing shouldNotHaveReceived could function without error.

### Why merge this feature add?
- Inspired by JetStream Teams and the possibility of being able to have a database per team. This is one step in that direction.
- The middleware recreates logic from deeper in this package, to best keep it in sync, I would like to see it apart of the upstream package.

#### Discussions this addresses:
- #389 
- #385 
- #378 
- #334 

#### Laracasts Discussions that seem to match the problem scope this PR is attempting to solve:
- https://laracasts.com/discuss/channels/general-discussion/multi-tenancy-with-multiple-databases
- https://laracasts.com/discuss/channels/general-discussion/app-architecture-feedback-multi-tenancy-multiple-data-centres
- https://laracasts.com/discuss/channels/general-discussion/multi-tenancy-with-separate-databases
- https://laracasts.com/discuss/channels/laravel/dynamic-multi-tenancy-best-practice
- https://laracasts.com/discuss/channels/general-discussion/is-my-application-considered-multi-tenancy

#### Incomplete internet tutorials
- https://mark.fullbrook.me/multi-tenant-databases-with-laravel-jetstream - At the end is specifically missing middleware code. - This is the first web page result when you Google for spatie multiteancy jetstream - https://www.google.com/search?q=Spatie+multitenancy+jetstream

### Requested Feedback
- Should I include more or less documentation? I was trying to strike a balance between showing the use of the middleware and giving them a step by step usage.
- Where would be good places to link to the new documentation. I didn't want to go around changing to much until this feature was overall deemed desirable.
- Middleware is currently reimplementing the same logic from the main `\Spatie\Multitenancy\Multitenancy::determineCurrentTenant` method, I'd love to reuse this method, however it's protected nature makes it inaccessible. Is there a recommended way to make these share code path? I'm especially interested in increasing the long term maintainability of the new middleware.
[...]

PR #399, opened 2022-10-14
Authentication Based Multi-Tenancy Documentation

kassah:
This PR changes the core `Multitenancy` service in ways required for authentication based multi-tenancy, as well as adding documentation on how to setup authentication aware multi-tenancy.

What changed?
In order to determineTenancy at a time other than application boot, I changed `Multitenancy::determineTenant` to public, and add the ability to allow the request to be passed in directly when available.

Why change `Multitenancy` service?
In previous PR, it was suggested that authentication aware determine tenant middleware be in documentation instead of as a class in the repository. In order to keep this class simple, and ensure it picks up future changes from this package, I wanted to reuse the determine tenant logic from where it is used by the rest of the package.

Are the changes tested?
I have added a test that ensures that the determine tenant method is public and does what is expected. Since the original service is not covered by the test suite, I did not have a good example to follow, so I tried to do my best to match other tests. I am open to feedback so that I can better match the project conventions.

Significant changes from previous PR:
- TenantFinder now uses Dependency Injection instead of Facade.
- DetermineTenant middleware uses Dependency Injection instead of Facade.
- Tests removed that covered the middleware, since it is now implemented in the documentation.

Dependency Change
Increased the minimum version of Mockery, so that tests with mocks utilizing shouldHaveReceived could function without error.

For reference of origin of this PR, see previous PR: https://github.com/spatie/laravel-multitenancy/pull/394

Credits
Thanks to @thomasruns for initial research into this subject and @lhartwell for feedback on wording, spelling and grammar with the additional documentation (just the docs addition, not this PR, errors here are my own!)

Current Questions I'm interested in getting answers to:
- [x] Is the advanced usage the best placement for this documentation.
- [x] Suggestions as to where this could be linked to from existing documentation.
- [x] Suggestions on how I might approach testing in a way that doesn't require a bump in the dependencies to resolve the mockery issue.
- [x] Should the documentation include samples with methods to add testing of their added code, or should this be left to be an exercise of the user? I already have the tests, just need to figure out if and how to include them.
- [x] This continues to have the issue of calling tenantFinder twice for every request, so possibly adding a conditional in the config that disables the determineTenant find at boot? thoughts?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     },
     "require-dev": {
         "laravel/legacy-factories": "^1.0.4",
-        "mockery/mockery": "^1.4",
+        "mockery/mockery": "^1.5",
         "orchestra/testbench": "^6.23|^7.0",
         "phpunit/phpunit": "^9.4",
         "spatie/valuestore": "^1.2",
--- a/docs/advanced-usage/authentication-aware-tenant-identification.md
+++ b/docs/advanced-usage/authentication-aware-tenant-identification.md
@@ -1,0 +1,149 @@
+---
+title: Authentication aware tenant identification
+weight: 10
+---
+
+Some projects will have their tenant based on a user attribute.
+
+On every request determining a tenant will happen at the very beginning of a request, even before routes and authentication are done. To utilize the tenantFinder after authentication, create and prioritize a middleware to determine tenant.
+
+## Base Installation
+
+Ensure the steps in [base installation](/docs/laravel-multitenancy/v2/installation/base-installation) have been completed with the fact that only some routes will be tenant aware in mind.
+
+## Create relation between users and tenant
+
+To identify a tenant from the User model, create a relation between the two.
+
+Create relation migration.
+
+```bash
+php artisan make:migration add_tenant_column_to_users
+```
+
+In the new migration file, create a relation between users and tenants.
+
+```php
+use Spatie\Multitenancy\Models\Tenant;
+
+return new class extends Migration
+{
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->foreignIdFor(Tenant::class)->nullable();
+        });
+    }
+
+    public function down()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->dropColumn('tenant_id');
+        });
+    }
+};
+```
+
+In users model, create eloquent relation indicating users belong to tenant.
+
+```php
+// in `app\Models\Users.php`
+use Spatie\Multitenancy\Models\Tenant;
+
+// ...
+
+    public function tenant()
+    {
+        return $this->belongsTo(Tenant::class);
+    }
+
+// ...
+```
+
+## Create custom tenant finder
+
+Create a tenant finder that returns the current tenant. Ensure it returns early if the user is not logged in.
+
+```php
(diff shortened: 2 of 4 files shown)


kassah:
@masterix21 @freekmurze Can you please review this PR so far, and provide feedback if this better fits the direction of the package? I also would appreciate feedback on the Questions above.

kassah:
@masterix21 Before incrementing versions, I recommend assessing which of [these ideas I should PR](https://github.com/spatie/laravel-multitenancy/discussions/categories/ideas?discussions_q=category%3AIdeas+author%3Akassah). You may want to have multiple of them in a single release. I am happy to pursue them as I have this PR. I just wanted to do this one first, as I knew it would help me know how to do the others so they required less review iterations.

kassah:
> @freekmurze, we should release a major version merging it because @kassah has changed the visibility to `src/Multitenancy.php`.

@masterix21 If you're following [semver](https://semver.org/), I believe this should be a minor version level change, as the change is backwards compatible. The only change I made was to make it public and add an optional argument. If called from within the class without an argument, it will continue to function exactly the same way.

freekmurze:
>  Is the advanced usage the best placement for this documentation.

Thanks for the detailed docs. However, I now think this might be best posted somewhere as a blog post, so we don't have to maintain it. We'd happily link to such blogpost in our docs.

> Suggestions on how I might approach testing in a way that doesn't require a bump in the dependencies to resolve the mockery issue.

I don't mind the dependencies being bumped

> This continues to have the issue of calling tenantFinder twice for every request, so possibly adding a conditional in the config that disables the determineTenant find at boot? thoughts?

I'd like to avoid having to add a flag for this in the base package. Maybe in your blog post, you can advise to add such a flag, a registering a custom tenant finder that respects that flag.


It may seem that I'm against this feature, but that's not the case. My concern is to keep the package as simple as it can be and minimise maintenance burden for us. I feel your use case is not a common one, and it can be achieve within the flexible setup of the package.

kassah:
@freekmurze I appreciate your responses.

> I'd like to avoid having to add a flag for this in the base package. Maybe in your blog post, you can advise to add such a flag, a registering a custom tenant finder that respects that flag.

I like this idea. I will look into integrating this into the instructions.

> I feel your use case [ .. ] can be achieve within the flexible setup of the package.

Without the modification of the main Multitnenacy class exposing the request processing (the only code change in this PR), I fear the code will be brittle and subject to future breakage as changes are applied to this package. Especially as that class is in one of the only untested areas of the package.

> My concern is to keep the package as simple as it can be and minimise maintenance burden for us.

After thinking about this PR further, with the discovery of the need to change the middleware priorities and no good API to modify that priority in Laravel 8 & 9 in a consistent and easily upgradable way. I actually agree, I think including these instructions might be too much to expect of this package.

> Thanks for the detailed docs. However, I now think this might be best posted somewhere as a blog post, so we don't have to maintain it. We'd happily link to such blogpost in our docs.

@freekmurze I don't currently have a blog. I'd be more interested in building a package that extends yours, as it's easier for me to maintain that with tests, etc. If I built that, would you still be up for a link to it being added to your documentation?

freekmurze:
Yes, if you fork the package, I'd happily link in our docs to your fork.

Closing this PR for now.

PR #404, opened 2022-10-21
Convert `Commands` tests to PEST

alexmanase:


masterix21:
@alexmanase, thanks for your time, but I think is better to keep only one test suite. So, if you like to help us feel free to convert all tests to pest. Thanks

alexmanase:
Sure! I will do it.

Issue #422, opened 2022-12-06
Current tenant model not refreshed in container before each job (stale model properties)


ionutantohi:
When a tenant is made current, there is a check to see if tenant we want to make current is already loaded. 

<img width="354" alt="image" src="https://user-images.githubusercontent.com/4609521/205852368-75c1586c-c5e8-47f1-a311-8175ff4544a1.png">

While this somehow makes sense, it **might** cause an issue in jobs, because jobs are processed by **long running processes**.

There was a back and forth around this check. Initially was added in #111 then removed in #115 and added back in 85c4e5c111c400f0db849302d56c22fbac21e9ce (cc @masterix21)

@telkins you asked in your pr #111 whether or not is a good idea. Let me give you an example.

Say we have the following tenant
id: 1
name: My tenant
**stripe_plan_id: abc**

Now, if we dispatch a job, for this tenant, then
- the worker will pick up the job
- there is no current tenant, then the tenant made current (loaded into container)
- job is processed

**Now, let's say that some info at tenant level are changed (by an user in his account etc)**

**stripe_plan_id: xyz**

When a second job is dispatched for the same tenant
- the worker will pick the job
- but since a tenant is already loaded into container and is the same (by looking at its id) it just returns early. But that means when calling `Tenant::current` inside the job, on second job `stripe_plan_id` is still `abc`. 

Note: The tenant is queried every time and before each job in `MakeQueueTenantAwareAction@listenForJobsBeingProcessed` but the object returned by `findTenant` (fresh from db) is never set to container because of id match.

Possible workarounds?
- remove the `isCurrent` check entirely, but I am not sure if that's ok. 
- update `isCurrent` check to also look at the tenant/model properties and return true only if all model properties are identical
- introduce a `forceMakeCurrent` method which should be called from `MakeQueueTenantAwareAction` which should bypass `isCurrent` check
- other?

masterix21:
Hi @ionutantohi, please create a repository to test the issue. Thanks

masterix21:
@freekmurze, one fix to this issue is to change the `MakeQueueTenantAwareAction.php` like so:

```php
class MakeQueueTenantAwareAction
{
    /// ...

    protected function listenForJobsBeingProcessed(): static
    {
        app('events')->listen(JobProcessing::class, function (JobProcessing $event) {
            $this->getTenantModel()::forgetCurrent();

            if (array_key_exists('tenantId', $event->job->payload())) {
[...]

PR #438, opened 2023-02-07
Update MakeQueueTenantAwareAction.php

masterix21:
The PR is a follow-up of issue #422

### DIFF
--- a/src/Actions/MakeQueueTenantAwareAction.php
+++ b/src/Actions/MakeQueueTenantAwareAction.php
@@ -41,13 +41,11 @@ class MakeQueueTenantAwareAction
     protected function listenForJobsBeingProcessed(): static
     {
         app('events')->listen(JobProcessing::class, function (JobProcessing $event) {
-            if (! array_key_exists('tenantId', $event->job->payload())) {
-                $this->getTenantModel()::forgetCurrent();
+            $this->getTenantModel()::forgetCurrent();
 
-                return;
+            if (array_key_exists('tenantId', $event->job->payload())) {
+                $this->findTenant($event)->makeCurrent();
             }
-
-            $this->findTenant($event)->makeCurrent();
         });
 
         return $this;
@@ -56,13 +54,11 @@ class MakeQueueTenantAwareAction
     protected function listenForJobsRetryRequested(): static
     {
         app('events')->listen(JobRetryRequested::class, function (JobRetryRequested $event) {
-            if (! array_key_exists('tenantId', $event->payload())) {
-                $this->getTenantModel()::forgetCurrent();
+            $this->getTenantModel()::forgetCurrent();
 
-                return;
+            if (array_key_exists('tenantId', $event->payload())) {
+                $this->findTenant($event)->makeCurrent();
             }
-
-            $this->findTenant($event)->makeCurrent();
         });
 
         return $this;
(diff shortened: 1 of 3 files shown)


masterix21:
@freekmurze, if your check is ok, I will merge the PR.

freekmurze:
Looks good to me 👍

masterix21:
Well, merged. Thanks.

moisish:
This PR has introduce a breaking change as per my bug here https://github.com/spatie/laravel-multitenancy/issues/446
Following the initial conversation https://github.com/spatie/laravel-multitenancy/issues/422 I believe changes on the Tenant model should be handle differently and not with the solution introduced in 3.0.2

masterix21:
@moisish, sorry, but using your logic, you will overwrite the current tenant if updates another one.

```php
$tenant1 = Tenant::find(1);
$tenant2 = Tenant::find(2);

$tenant1->makeCurrent();

$tenant2->fill(['host' => 'example-two.test'])->save();

echo Tenant::current()->id; // will print 2
```

moisish:
@masterix21 You are right. We are still facing the issue tho so we need to find a solution

danijelk:
Adding our issue to the list as well.

We use 'queues_are_tenant_aware_by_default' => false, and everything was fine. In 3.0.2 all jobs forget current tenant and the rest of the assertions fail as it forgets tenant for the rest of the code running after a dispatch.

Setting it to true it will forget and re-set it, but once any dispatch_now is done the makeCurrent doesn't seem to reconnect - tenant database is now empty, I am yet to find the root cause.


Looking at the original problem I see why this was changed, I would have preferred an interface _JobShouldRefreshTenant_ which would trigger a get/forget when checking with _implementsInterface_. This is cause we never change any tenant data and would like to avoid a forget/reset overhead if there is no reason to do it.

benddailey:
I believe #438 should be reverted based on the reports of many. I found this thread while dealing with failed Broadcast and failed laravel-medialibrary conversion jobs. I realized the issue is caused because of the long running worker queue. I ended up on rather simple fix for this issue. Switch production to run worker:listen instead of worker:queue. There is a very small performance penalty for booting the framework for each job but it is better to get the right and proper result than having failed jobs or broadcast messages that don't go anywhere because the have the wrong config parameters.

I switched my supervisor configuration for my worker /etc/supervisor/conf.d/wildcard_auctionit_io-worker.conf as seen below:

```
[program:wildcard_auctionit_io-worker]
process_name=%(program_name)s_%(process_num)02d
#command=/usr/bin/php /var/www/wildcard_auctionit_io/artisan queue:work --sleep 1 --timeout 120
command=/usr/bin/php /var/www/wildcard_auctionit_io/artisan queue:listen --sleep 1 --timeout 120
numprocs=4
autostart=true
autorestart=true
user=www-data
redirect_stderr=true
stdout_logfile=/var/www/wildcard_auctionit_io/storage/logs/worker.log
```
Maybe the documentation should be updated to mention running queue:listen instead of queue:work?

mbardelmeijer:
I agree that #438 should be reverted and moved to a major release (with additional fixes). We just spent a few hours debugging an edge case that also fell back to this. 

I agree that the changes made should still be relevant in an upcoming major release but seem to have breaking changes on a minor release.

PR #452, opened 2023-03-13
Fix typo in comment in multitenancy.php

inmanturbo:
This tiny patch adds a single letter to a comment in the config file to fix a tiny typo :)


inmanturbo:
Going to resubmit with [no ci] in commit message. No sense running ci for fixing a non decorator typo in a config comment.

PR #454, opened 2023-03-13
[no ci] Fix typo in config comment

inmanturbo:
The comment means to say "your" but instead is missing the "y" and says "our".


masterix21:
Thanks

PR #459, opened 2023-03-22
Fix unmatched double quote (typo)

inmanturbo:
The value that the `tenant` option is set to is followed by a double quote, but it does not have an opening double quote.


masterix21:
Thanks

PR #462, opened 2023-04-03
Update TenantCollection.php

1cbyc:
Okay, this is it. I used the 'use' keyword to import the 'Tenant' class at the top of the file, instead of using the fully qualified namespace in the code. Then, I used the 'Collection::map' method instead of 'Collection::each' method to iterate over the items and return a new collection with the modified items. I also renamed the '$callable' parameter to '$callback' to make it even more clear.


masterix21:
Hi @1cbyc, thanks for your time on it, but I prefer our current code.

PR #463, opened 2023-04-03
Update MultitenancyServiceProvider.php

1cbyc:
It's safer to use the 'Illuminate\Events\Dispatcher' class instead of the 'Illuminate\Support\Facades\Event' facade though. I also think that instead of using the 'isset()' function to check if the 'LARAVEL_OCTANE' environment variable is set, use the 'env()' function to retrieve its value. Considered using the 'app()->singleton()' method to register the 'Multitenancy' class as a singleton, instead of using '$this->app->bind()' in the 'packageBooted()' method. Actually tried something you might not consider necessary too; Renaming the 'packageBooted()' method to 'booted()' to conform to the naming convention of Laravel service providers.

### DIFF
--- a/src/MultitenancyServiceProvider.php
+++ b/src/MultitenancyServiceProvider.php
@@ -2,7 +2,7 @@
 
 namespace Spatie\Multitenancy;
 
-use Illuminate\Support\Facades\Event;
+use Illuminate\Events\Dispatcher;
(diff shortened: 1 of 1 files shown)


masterix21:
Hi @1cbyc, thanks for your time on it, but I prefer our current code.

Issue #467, opened 2023-04-20
Current tenant not determined when dispatching queued closures


hanicab:
Hello,

Current tenant could not be determined when dispatching queued closure from tenant context, even if jobs are configured to be tenant aware by default.

```
$podcast = App\Podcast::find(1);
 
dispatch(function () use ($podcast) {
    $podcast->publish();
});
```

Can you please help to resolve this problem ?

How to make queued closure tenant aware ?

Best regards

benddailey:
@hanicab How are you running your queue workers?

I found that a similar issue is caused because of the long running worker queue. I ended up on rather simple fix for this issue. Switch production to run worker:listen instead of worker:queue. There is a very small performance penalty for booting the framework for each job but it is better to get the right and proper result.

hanicab:
@benddailey Thank you for your reply,

I use horizon,

The problem is only when dispatching queued closure from tenant context (as described)

```
MyCustomJob::dispatch(); //works perfectly

dispatch(function(){
   Order::where('id',123)->delete();
}); // not working
```

masterix21:
Try again with the release 3.0.3. Thanks!

bcalik:
> Try again with the release 3.0.3. Thanks!

I'm on 3.0.3 and I encounter the same error.

masterix21:
@bcalik, please try adding `CallQueuedClosure::class` to `queueable_to_job` in `config/multitenancy.php`, so finally, you should have the config like so:
```php
'queueable_to_job' => [
        SendQueuedMailable::class => 'mailable',
        SendQueuedNotifications::class => 'notification',
        CallQueuedClosure::class => 'closure', // <-- ADD THIS
        CallQueuedListener::class => 'class',
        BroadcastEvent::class => 'event',
    ],
```

Thanks

masterix21:
Solved with release 3.1.0.

PR #492, opened 2023-09-25
Add support to queued closures

masterix21:
It adds support to queued closures with tests.

Issue: https://github.com/spatie/laravel-multitenancy/issues/467


masterix21:
@freekmurze, do you happen to have an idea about the test failures?

freekmurze:
Not really, but if the lower deps fail, I usually just bump to lower requirements to something higher.

PR #495, opened 2023-11-08
Update migration stub

sevannerse:
It makes the structure of migration up-to-date.


masterix21:
Thanks for your time.

PR #502, opened 2023-12-03
Add new `TenantNotFoundForRequestEvent`

jacobmllr95:
This PR adds a new `TenantNotFoundForRequestEvent` which get's fired when no tenant was found by the `findForRequest()` method of the `TenantFinder`.

While it is possible to write your own `TenantFinder` and implement the behavior yourself, I think this is a good addition to have out of the box for people who don't want to go this extra step.

The event itself can be useful for scenarios where a tenant could not be found for the current subdomain and you wan't to redirect to e.g. the root domain.


masterix21:
Would you describe a real example case?

jacobmllr95:
@masterix21 Assumed you have an application where each tenant has it's own subdomain and the Laravel application handles the traffic for all wildcard subdomains except some known subdomains like `www.yourtenancyapp.com`.

This feature would come in handy when e.g. a request for `unknown.yourtenancyapp.com` comes in and there is no matching tenant for that namespace and you would like to redirect to the root domain in that case.

The event listener in that case could be something like:
```php
<?php

namespace App\Listeners;

use Illuminate\Support\Facades\Redirect;
use Spatie\Multitenancy\Events\TenantNotFoundForRequestEvent;

class TenantNotFoundForRequestEventListener
{
    /**
     * Handle the event.
     *
     * @param TenantNotFoundForRequestEvent $event
     * @return void
     */
    public function handle(TenantNotFoundForRequestEvent $event): void
    {
        // When no tenant could be resolved by request, redirect to the root domain.
        Redirect::away('www.yourtenancyapp.com')->send();
        exit();
    }
}
```

masterix21:
Thanks!

newtonjob:
@jacobmllr95 Although this is already merged, I wouldn't recommend sending a redirect response from an event listener and exiting the request flow right there, as it doesn't allow the response to go out normally through the Laravel request pipeline.

It makes more sense to do this kind of stuff in a middleware. 
```php
    public function handle($request, Closure $next)
    {
        if (! Tenant::checkCurrent()) {
            return redirect()->away('www.yourtenancyapp.com');
        }

        return $next($request);
    }
```

Alternatively, the `NeedsTenant` middleware that comes out of the box will throw a `NoCurrentTenant` exception and you can handle this in your `App\Exceptions\Handler` to return a redirect response.

```php
   /**
     * Register the exception handling callbacks for the application.
     */
    public function register(): void
    {
        $this->renderable(function (NoCurrentTenant $e) {
            return redirect()->away('www.yourtenancyapp.com');
        });
    }
```

PR #503, opened 2023-12-04
Show tenant info for all `TenantAware` commands

jacobmllr95:
This PR enables the output of the tenant info for each command that uses the `TenantAware` trait, instead of being only available for the `TenantsArtisanCommand`.

### DIFF
--- a/src/Commands/Concerns/TenantAware.php
+++ b/src/Commands/Concerns/TenantAware.php
@@ -29,9 +29,19 @@ trait TenantAware
             return -1;
         }
 
-        return $tenantQuery
-            ->cursor()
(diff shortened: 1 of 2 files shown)


masterix21:
Thanks for your time on it, but I prefer to keep the command executions as clean as possible: it allows you to see the information that interests you command by command.

jacobmllr95:
@masterix21 Fair point.

Would making it optional be an option for you?
IMHO it's a functionality that should be available in the `TennatAware` trait and it makes the `TenantsArtisanCommand` very clean and simple on the other hand.

masterix21:
I think no because it introduces an unnecessary complexity.

PR #540, opened 2024-06-25
Add tenant caching

francoism90:
We've noticed the database is hit a lot, because we use a lot of (sub)domains for campaigns and such.

This will add a tenant cache, and reduce the time and load to serve the requested tenant. I have also added a fallback, just to make the tenant can be reached.

Please let me know any improvements or changes when needed. :)


masterix21:
Thanks for your time on it. I think could be mergeable but requires some improvements.

Anyway, I prefer to remand the decision to @freekmurze: If he likes the feature introduced by your PR, I will review it.

francoism90:
@masterix21 Thanks for your consideration. :)

If you could tell me what improvements you will like to see, I could potentially add them or make changes.

freekmurze:
I appreciate your work on this, but to keep the package as small and simple as can be, I'm going to pass on this for now.

I feel that for the people that are in this scenario, they can manually add caching (or whatever optimizations they want, in a custom `TenantFinder`. I would be open to a page in the docs that explains how to set this up though.

PR #544, opened 2024-07-18
Contract IsTenant replace Tenant uses

masterix21:
PR follows the discussion here: https://github.com/spatie/laravel-multitenancy/discussions/543


masterix21:
@freekmurze, I'm ready for your verification. I'm in love with these changes.

freekmurze:
Looks good! Maybe also drop support for Laravel 10 and provide upgrade instructions in `UPGRADE.md` 👍 

Thanks for your work on this.

masterix21:
I'm ready. Can I merge?

freekmurze:
Do you have other breaking changes in mind? Now would be the time to implement those too.

masterix21:
I think not. I changed a few things in the tests, so apart from what I mentioned in the upgrade guide, it currently seems that there are no other breaking changes.

masterix21:
A possible idea could be to use the `Context` instead of the service container with the currentTenant.

https://laravel.com/docs/11.x/context#main-content

benddailey:
I think using the new context feature could be a really great idea especially since you are dropping Laravel 10 support and releasing a new major version. I remember coming upon this tweet shortly after context was released. Maybe it can provide inspiration? https://twitter.com/davidhemphill/status/1770893521748123683 Thanks @davidhemphill

masterix21:
I have made some test using context, but when you make a tenant as current and then call `Context::add('tenantId', 1)`, every job will contain the context if it is NotTenantAware.

We can reduce our code, but we should remove our `TenantAware` and `NotTenantAware` from the jobs. The only advantage is that it will help with all jobs/events dispatched by other packages. 

@freekmurze: What is your opinion about that?

freekmurze:
Using context seems like the way to go. Let's make the context key,  `tenantId`, configurable.

> every job will contain the context if it is NotTenantAware.

That's no problem I think

> We can reduce our code, but we should remove our TenantAware and NotTenantAware from the jobs. The only advantage is that it will help with all jobs/events dispatched by other packages.

I don't understand why this should be removed, could you give some more details on that?

freekmurze:
Ready again for review?

masterix21:
Yes

freekmurze:
It all looks very good, thanks for your work on this!

PR #572, opened 2024-12-16
Fix sqlite driver database lock

Haruki1707:
This pull request fixes an issue with the `tenants:artisan` command when using the SQLite driver. Previously, running commands like `tinker` would lock the database file because the cursor remained open, preventing the connection from closing until all tenants were processed.

The fix checks if the `IsTenant`-defined model is using a connection with the SQLite driver. If so, it retrieves the tenants with a `get()` instead of using a `cursor()`, ensuring the connection is closed promptly and the database file is not locked.

### DIFF
--- a/src/Commands/Concerns/TenantAware.php
+++ b/src/Commands/Concerns/TenantAware.php
@@ -28,8 +28,10 @@ trait TenantAware
             return -1;
         }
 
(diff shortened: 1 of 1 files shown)


masterix21:
Currently, we only test the package with MySQL. To merge the PR, we should be sure that the other parts are okay with SQLite. I appreciate your work, but this is not the time to integrate it. Thanks

PR #586, opened 2025-04-21
A new way to use a model as a Tenant and Landlord.

mckaragulle:
Option 3: Using a model as both tenant and landlord

It automatically detects when logged in as a `Tenant` or `Landlord`. It automatically sets the database connection name.

With this method, the database connection error that occurs when matching roles and permissions while using the `Laravel spatie/permission ` package is also prevented.

With this method, you do not need to use different Models that use the same table name on the `Landlord` and `Tenant` sides.

You can use this instead of `Option 2`.

To accomplish that use trait `UsesBothConnection` 

Here's an example:

```php
namespace App\Models;

use Illuminate\Database\Eloquent\Factories\HasFactory;
use Spatie\Permission\Models\Role as SpatieRole;
use Spatie\Multitenancy\Models\Concerns\UsesBothConnection;

class Role extends SpatieRole
{
    use HasFactory;
    use UsesBothConnection;
}
```


masterix21:
Thanks for your time, but it may introduce confusion. I prefer to keep the package clean.

Thanks

Issue #575, opened 2024-12-29
Issue decrypting jobs that implement the ShouldBeEncrypted contract.


scotchdesign:
I had no issues with v3.

Since upgrading to v4 I noticed that jobs that implement the ShouldBeEncrypted contract now fail, throwing an error when being processed:
unserialize() error at offset 0 of XX bytes

Digging deeper, I can trace this back to Spatie's IsTenantAware function. It appears the payload is still encrypted when IsTenantAware tries to unserialize it, so the unserialize fails (throwing the error above).

Again, wasn't an issue with v3, is an issue with v4, and haven't had a chance to dig deeper into what has changed in your codebase between the versions.

francoism90:
@skoontastic Hopefully you solved this already.

If not, are you using redis + compression? I had an issue with this.

masterix21:
Please post the code that throws the exception to solve the issue. Thanks

PR #588, opened 2025-05-08
[Bugfix]: Jobs with `ShouldBeEncrypted` need to be decrypted

Joel-Jensen:
Hi, 

Minor fix for jobs with the interface `Illuminate\Contracts\Queue\ShouldBeEncrypted`, which fixes my issue when running https://github.com/nWidart/laravel-modules and the same issue in https://github.com/spatie/laravel-multitenancy/issues/575.

**Error:**
`unserialize() error at offset 0 of XX bytes`

**Background:**
The interface will encrypt the `$payload['data']['command']` value and is therefore required to be decrypted before we try to `unserialize` it. Below is the Laravel code that encrypts the value.
```php
$command = $this->jobShouldBeEncrypted($job) && $this->container->bound(Encrypter::class)
    ? $this->container[Encrypter::class]->encrypt(serialize(clone $job))
    : serialize(clone $job);
``` 
https://github.com/laravel/framework/blob/12.x/src/Illuminate/Queue/Queue.php#L170

The solution to check if it's encrypted is taken from this Laravel implementation:
https://github.com/laravel/framework/blob/12.x/src/Illuminate/Queue/Console/RetryCommand.php#L183 

As mentioned in the issue, this wasn't an issue in v3, since `bindOrForgetCurrentTenant` in the event listeners `app('events')->listen(JobProcessing::class, ...)` never used the `$payload['data']['command']`. It was only used in the queue binding `app('queue')->createPayloadUsing(...)`.
https://github.com/spatie/laravel-multitenancy/blob/v3/src/Actions/MakeQueueTenantAwareAction.php

### DIFF
--- a/src/Actions/MakeQueueTenantAwareAction.php
+++ b/src/Actions/MakeQueueTenantAwareAction.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\Multitenancy\Actions;
 
+use Illuminate\Contracts\Encryption\Encrypter;
 use Illuminate\Queue\Events\JobProcessing;
 use Illuminate\Queue\Events\JobRetryRequested;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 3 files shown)


masterix21:
Thanks for the effort.

PR #596, opened 2025-05-27
Add types to TenantCollection

markieo1:
Adds the generic types to the `TenantCollection`. This can be used in projects that use PHPStan with Type Checking enabled.


masterix21:
Thanks

PR #601, opened 2025-08-19
TenantAware `before` Method

SethSharp:
When using the `TenantAware` trait, I encountered unexpected behaviour where a global variable retained its value between tenant executions. My command was keeping count of a specific value (starting at 0), but when the next tenant command executed, it was not 0 but the count from the previous execution.

I came up with this solution to implement a `before` method, which developers can hook into to fluently define any logic that should be run before each tenant execution. I'm on the fence with this as its probably not necessary when most of the time its easy enough to add this at the start of each command, but wouldn't it be nice to have a method that automatically applies before each execution without adding it to the start of the command?

Happy to hash this out or just update the documentation to highlight this behaviour.

### DIFF
--- a/docs/advanced-usage/making-artisan-commands-tenant-aware.md
+++ b/docs/advanced-usage/making-artisan-commands-tenant-aware.md
@@ -65,3 +65,12 @@ If the command only needs to run for a specific tenant, you can pass its `id` to
 ```bash
 php artisan tenants:artisan "migrate --seed" --tenant=123
 ```
+
+### Using `before`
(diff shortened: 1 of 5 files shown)


SethSharp:
Hey @masterix21 thanks for checking it out!

I think the main benefit here is convenience, not requiring the developer to remember to call it at the start. It also matches Laravel and other packages (like `boot` or `mount` hooks in Livewire), which are handled under the hood.

The manual approach has one main benefit, control of when its called, but it comes down to the use case. I think `before` has a place in this package as it gives devs a fluent solution to handling the state persistence issue.

PR #602, opened 2025-08-20
Command state persistence with `TenantAware`

SethSharp:
When using the `TenantAware` trait, I encountered unexpected behaviour where a global variable retained its value between tenant executions. My command was keeping count of a specific value (starting at 0), but when the next tenant command executed, it was not 0 but the previous tenant count.

I have updated the documentation to highlight this behaviour to future developers.


masterix21:
Thanks

PR #607, opened 2025-09-24
feat: Add configuration option to specify `TenantAware` and `NotTenantAware` interfaces

przemyslaw-przylucki:
The goal of this PR is to add 2 new keys in the config file that'll allow the user to specify the interface that determines if given job is TenantAware.

The usecase for this is as follows:

1. I'm building a "foundation" package to be used by multiple consumers that themselves use the spatie/laravel-multitenancy
2. I do not want the "foundation" package to have this package installed as dependency
3. Without going into much detail, the "foundation" package provides abstract, basic stuff like BaseJob, or SchedulerJob, where I want to specify some tenancy behavior, those classes will be extended on the app-level, but I don't want consumers to always add `implements NotTenantAware`

Alternative solutions:
1. Without those interfaces being configurable, I need to override `MakeQueueTenantAwareAction` (or at least function on it `isTenantAware`), which is completely acceptable solution if you do not believe that this configuration option benefits the broader public
2. I can modify this so that the `tenant_aware_jobs` and `not_tenant_aware_jobs` would instead also validate all the parents of given job class, which you could argue makes it easier to specify a couple of different inheritance exclusions

Thanks for this beautiful package :)

### DIFF
--- a/config/multitenancy.php
+++ b/config/multitenancy.php
@@ -1,8 +1,10 @@
 <?php
 
+use Spatie\Multitenancy\Jobs\TenantAware;
 use Illuminate\Broadcasting\BroadcastEvent;
 use Illuminate\Events\CallQueuedListener;
 use Illuminate\Mail\SendQueuedMailable;
+use Spatie\Multitenancy\Jobs\NotTenantAware;
 use Illuminate\Notifications\SendQueuedNotifications;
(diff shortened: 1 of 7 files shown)


masterix21:
Hi @przemyslaw-przylucki, thanks for your effort.

PR #608, opened 2025-09-26
Fix syntax for tenant aware interface config

danharrin:
The config options are attempting to be run as shell commands, as backticks are being used instead of quotes. Introduced #607.

https://www.php.net/manual/en/language.operators.execution.php

@masterix21 this is quite a major issue introduced in today's release.


masterix21:
Thanks for your time on it.

PR #609, opened 2025-10-03
Automatically set database.default to tenant connection in SwitchTenantDatabaseTaskUpdate SwitchTenantDatabaseTask.php

vahidalvandi:
Currently, when switching tenants using SwitchTenantDatabaseTask, the tenant connection is configured
and purged, but the Laravel default database connection (`database.default`) remains pointing to 
the landlord connection. This means that any query or model that doesn't explicitly specify the tenant 
connection will continue to run on the landlord database.

This PR updates the setTenantConnectionDatabaseName() method to also set the default connection 
to the tenant connection when switching tenants. This ensures that:

- All queries and Eloquent models use the correct tenant database by default.
- The landlord connection is still available for models that explicitly specify it.
- The change does not affect ForgetCurrentTenantAction, which will reset the default connection 
  back to the landlord when the tenant is forgotten.

Example:

    config([
        'database.connections.tenant.database' => $databaseName,
        'database.default' => 'tenant',  // Automatically switch default
    ]);

This makes multi-tenant setups more intuitive and prevents accidental queries to the landlord database.

### DIFF
--- a/src/Tasks/SwitchTenantDatabaseTask.php
+++ b/src/Tasks/SwitchTenantDatabaseTask.php
@@ -36,6 +36,7 @@ class SwitchTenantDatabaseTask implements SwitchTenantTask
 
         config([
             "database.connections.{$tenantConnectionName}.database" => $databaseName,
+            "database.default" => $tenantConnectionName,
         ]);
 
         app('db')->extend($tenantConnectionName, function ($config, $name) use ($databaseName) {


vahidalvandi:
if default connection is not tenancy

masterix21:
Hi @vahidalvandi, there have been numerous discussions, issues, and PRs about this, so it is not the first time. 

We prefer to keep things as they are, because the default connection should always be known and ready to use.

Thank you for your time on this.

vahidalvandi:
Even though you said it was discussed before, SwitchTenantDatabaseTask alone cannot change "database.default" with the current code.
In practice, if the model does not explicitly use use UsesTenantConnection; , the task switch database will not work and in many cases, not all models can be changed.

PR #627, opened 2026-02-13
Add caching layer for TenantFinder

moisish:
## Summary                                                                                                                                                                         
                                         
  Tenant resolution via `TenantFinder` hits the database on every request. This PR adds an opt-in caching decorator that wraps any `TenantFinder` implementation and caches the resolved tenant.                                                                                                                                                                   
                                                                                                                                                                                     
  ## What's included                                                                                                                                                                 

  - **`CachingTenantFinder`** — a decorator that wraps the configured `TenantFinder`, caches the result, and delegates to the inner finder on cache miss. Supports configurable TTL, cache store, and forever-cache (`ttl: 0`). Null results are intentionally not cached.
  - **`CachesTenantFinderResults`** — interface exposing `forget(Request)` and `forgetByKey(string)` for cache invalidation, allowing other packages to type-hint without depending
  on the concrete class.
  - **Config-driven** — disabled by default. Users enable it and optionally swap the class for custom cache key strategies (e.g. subdomain-based resolution):

  ```php
  'tenant_finder_cache' => [
      'enabled' => false,
      'class' => CachingTenantFinder::class,
      'store' => null,
      'ttl' => 300, // seconds, 0 = forever
  ],
  ```

  - Extensible — override resolveCacheKey() or cachePrefix() for custom cache key derivation:
 ```php
  class SubdomainCachingTenantFinder extends CachingTenantFinder
  {
      protected function resolveCacheKey(Request $request): string
      {
          return $this->cachePrefix() . explode('.', $request->getHost())[0];
      }
  }
  ```

### DIFF
--- a/src/Contracts/InvalidatesTenantCache.php
+++ b/src/Contracts/InvalidatesTenantCache.php
@@ -4,7 +4,7 @@ namespace Spatie\Multitenancy\Contracts;
 
 use Illuminate\Http\Request;
 
-interface CachesTenantFinderResults
+interface InvalidatesTenantCache
 {
     public function forget(Request $request): bool;
 
--- a/src/TenantFinder/CachingTenantFinder.php
+++ b/src/TenantFinder/CachingTenantFinder.php
@@ -4,10 +4,10 @@ namespace Spatie\Multitenancy\TenantFinder;
 
 use Illuminate\Cache\Repository;
 use Illuminate\Http\Request;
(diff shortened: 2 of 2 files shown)


masterix21:
Hey @moisish, thanks for the PR! 

Since TenantFinder is fully swappable via config, you can add a caching layer in your own implementation without needing changes to the package itself — keeping things flexible per project. Also, the test suite is currently failing, so we can't merge this as-is.

Thanks again for the contribution, we appreciate it!

PR #629, opened 2026-02-20
Tenant not determined during browser testing

moisish:
### The Problem

When running Pest browser tests, the current tenant is never determined. This is because
`Multitenancy::configureRequests()` checks `$this->app->runningInConsole()` and skips `determineCurrentTenant()` entirely when running in a console context.

Since the test runner executes in the console, tenant resolution is bypassed — even though the application is handling
real HTTP requests during browser tests.

### My Workaround

To work around this in my project, I had to register a custom middleware in my `TestCase::setUp()` that manually
resolves the tenant for every request:

```php
protected function setUp(): void
{
    parent::setUp();

    $this->app->bind('pest-resolve-tenant', fn () => new class
    {
        public function handle(Request $request, Closure $next): Response
        {
            if (! app(IsTenant::class)::checkCurrent()) {
                app(TenantFinder::class)->findForRequest($request)?->makeCurrent();
            }

          return $next($request);
      }
  });

  $this->app['router']->prependMiddlewareToGroup('web', 'pest-resolve-tenant');
}
```

This works but is boilerplate that every user running browser tests would need to add.

### Proposed Fix

Add a new config option determine_current_tenant_in_tests (default false) that allows tenant determination when running in a test context.
It only applies during tests and is fully opt-in — no change in default behavior.
`src/Multitenancy.php — Updated configureRequests():`

```php
protected function configureRequests(): static
{
    if (! $this->app->runningInConsole() || ($this->app->runningUnitTests() && $this->app['config']->get('multitenancy.determine_current_tenant_in_tests'))) {
        $this->determineCurrentTenant();
    }

    return $this;
}
```

### DIFF
--- a/tests/Feature/MultitenancyTest.php
+++ b/tests/Feature/MultitenancyTest.php
@@ -1,0 +1,57 @@
+<?php
+
+use Spatie\Multitenancy\Events\TenantNotFoundForRequestEvent;
+use Spatie\Multitenancy\Models\Tenant;
+use Spatie\Multitenancy\Multitenancy;
+use Spatie\Multitenancy\TenantFinder\DomainTenantFinder;
+
+it('does not determine tenant in console by default', function () {
+   Tenant::factory()->create(['domain' => 'example.com']);
+
+    config()->set('multitenancy.tenant_finder', DomainTenantFinder::class);
+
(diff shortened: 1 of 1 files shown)


moisish:
This is an issue that other people will have. The default behaviour does not change so I disagree with the unnecessary complexity.

PR #633, opened 2026-04-09
Add description to TenantsArtisanCommand

### LINKED ISSUE #632: Method "Symfony\Component\Console\Command\Command::getDefaultDescription()" is deprecated
Because this class:

https://github.com/spatie/laravel-multitenancy/blob/main/src/Commands/TenantsArtisanCommand.php

Is missing `protected $description = 'Some text';`

And, if I have the misfortune of passing through:

https://github.com/laravel/framework/blob/3ff39b7a9b83e633383ec9b019827ed54b6d38bc/src/Illuminate/Console/Command.php#L106

I end up with:

```
 Method "Symfony\Component\Console\Command\Command::getDefaultDescription()" is deprecated
```

In my case, it's poluting my phpunit/pest tests.

Please add some sort of `protected $description` to the command?

Thank you for your consideration.

dac514:
Add a default description

Fixes: https://github.com/spatie/laravel-multitenancy/issues/632#issuecomment-4214380470

## More info

Other projects have had this issue too, for example: https://github.com/DirectoryTree/LdapRecord-Laravel/issues/689

Laravel fixed it upstream, so this only affects projects that aren't up-to-date in their point/patch releases, but it's still good to have a default IMO, see: https://github.com/laravel/framework/pull/55888

### DIFF
--- a/src/Commands/TenantsArtisanCommand.php
+++ b/src/Commands/TenantsArtisanCommand.php
@@ -15,6 +15,8 @@ class TenantsArtisanCommand extends Command
 
     protected $signature = 'tenants:artisan {artisanCommand} {--tenant=*}';
 
+    protected $description = 'Run an Artisan command for selected tenants.';
+
     public function handle(): void
     {
         if (! $artisanCommand = $this->argument('artisanCommand')) {


masterix21:
Thanks

PR #634, opened 2026-04-21
Check if current session is available

francoism90:
Sometimes you use multitenancy on API-routes. These can have a session (e.g. with sanctum), and sometimes don't (depending on the route as well). Using the middleware on api routes, does offer some protection layer.

On Fortify they https://github.com/laravel/fortify/blob/fc66a38872a0e304748336d2975f37672c8fca9c/src/Http/Controllers/AuthenticatedSessionController.php#L104 check if a session exists. I think this can be the same for this package, but use an early return instead.

I also added the key as config option. Would that be a breaking change? Would you want a fallback key? The reason is that it allows the user to configure this key, like with other keys as well.

Let me know what you think. :)

### DIFF
--- a/config/multitenancy.php
+++ b/config/multitenancy.php
@@ -79,6 +79,12 @@ return [
     'current_tenant_container_key' => 'currentTenant',
 
     /*
+    * This key will be used to store the tenant id in the session when using the
+    * `EnsureValidTenantSession` middleware.
+    */
+    'current_tenant_session_key' => 'ensure_valid_tenant_session_tenant_id',
+
+    /*
      * Set it to `true` if you like to cache the tenant(s) routes
      * in a shared file using the `SwitchRouteCacheTask`.
      */
--- a/src/Concerns/UsesMultitenancyConfig.php
+++ b/src/Concerns/UsesMultitenancyConfig.php
@@ -27,6 +27,11 @@ trait UsesMultitenancyConfig
         return config('multitenancy.current_tenant_container_key');
     }
 
+    public function currentTenantSessionKey(): string
+    {
+        return config('multitenancy.current_tenant_session_key');
+    }
+
     public function getMultitenancyActionClass(string $actionName, string $actionClass)
     {
         $configuredClass = config("multitenancy.actions.{$actionName}") ?? $actionClass;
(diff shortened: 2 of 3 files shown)


masterix21:
Hi @francoism90, thanks for the PR and for taking the time to explain your reasoning!

I've thought it through and I'd like to share why I'm leaning towards not merging this, so we can discuss.

**On the `hasSession()` bypass**

The middleware is called `EnsureValidTenantSession` — its whole purpose is to guarantee that a session cannot be reused across tenants. If we silently skip requests that don't have a session, we change its contract from "enforces isolation" to "enforces isolation when convenient", and a misconfiguration (e.g. applying it to a stateless route group by mistake) would no longer surface as an error — it would silently grant access. I'd rather the middleware fail loudly than pass through.

The Fortify reference isn't quite equivalent: there `hasSession()` is used to make logout idempotent, not to gate a security check.

**On the API use case**

In practice, the middleware already works on API routes whenever a session is actually present — for example, Sanctum SPA authentication uses the session cookie, so the middleware behaves correctly there. For truly stateless API routes (token-only, no session), `NeedsTenant` on its own is the right tool; `EnsureValidTenantSession` was never meant for that scenario, since there's no session to validate against.

So the pattern I'd suggest is: apply `EnsureValidTenantSession` only to route groups that actually carry a session, and use `NeedsTenant` alone for stateless endpoints. If a project has a mixed group, a thin app-level wrapper is cleaner than a package-level bypass.

**On the configurable session key**

I don't see a concrete use case where users would need to change this key — it's an internal implementation detail, and adding a config option (plus a potential breaking change for users with a published config file) doesn't seem to buy us much. Happy to revisit if you have a scenario in mind where the default name actually causes friction.

Given the above, I'm going to close this one, but I really appreciate the discussion — the question about middleware semantics on API routes is a good one, and worth calling out more clearly in the docs. I'll make a note to improve that section.

Thanks again! 🙏

francoism90:
@masterix21 Thanks for explaining why/how it should already work on the API (Sanctum in this case), and making a note to improve the docs. :)

After your giving examples, I also think it makes more sense to define it per controller/route for API routes, instead of forcing it globally. It indeed would also be possible to expand `EnsureValidTenantSession` middleware class itself.

Appreciate your answer as well!

PR #640, opened 2026-07-08
Support legacy queue payloads in tenant-awareness detection

ragulka:
### Problem

`MakeQueueTenantAwareAction::isTenantAware()` reads `$payload['data']['command']` unconditionally. Modern Laravel payloads always carry the serialized job there, but **legacy string-job payloads have no `data.command`**:

```json
{ "job": "App\\Jobs\\SyncInventory@handle", "data": { "sku": "ABC-123" } }
```

This shape is produced whenever messages reach the queue from something other than Laravel's dispatcher — e.g. **AWS EventBridge input transformers** pushing to SQS, or any raw `Queue::pushRaw`-style producer. Laravel's worker still runs them via `$instance->handle($job, $data)`, and fires `JobProcessing` as usual.

Because `data.command` is absent, the action throws **`Undefined array key "command"`** before it ever reaches the interface checks. So a legacy-format job that implements `NotTenantAware` (or `TenantAware`) has its declaration silently ignored, and the worker errors instead of processing the job.

### Fix

When `data.command` is absent, resolve the job class straight from `payload['job']` (stripping the `@method` suffix) and run the **existing** interface / config-list / default checks against that class name (`ReflectionClass` accepts a class-string). The decision logic is extracted into a shared helper so both paths behave identically.

- Honors all the same config knobs: `tenant_aware_interface`, `not_tenant_aware_interface`, `tenant_aware_jobs`, `not_tenant_aware_jobs`, `queues_are_tenant_aware_by_default`.
- If the class can't be resolved (missing `job` key, or class doesn't exist) it falls back to `queues_are_tenant_aware_by_default` — it never throws.
- The legacy path has nothing to unserialize and no encrypted command, so no decryption/unserialization is attempted there.

### Zero impact on modern payloads

Modern `Queueable` jobs always carry `data.command` (the real job is serialized inside it; the top-level `job` is `Illuminate\Queue\CallQueuedHandler@call`, not the job class). The new branch triggers **only** when `data.command` is genuinely absent, so the modern path is unchanged — the added risk surface is limited to payloads that currently crash. A regression test asserts this and fails if the legacy branch is ever allowed to touch a modern payload.

### Tests

New `LegacyQueuePayloadTest` covering: legacy `NotTenantAware` → false; legacy `TenantAware` → true; legacy class in `tenant_aware_jobs` / `not_tenant_aware_jobs` → respective result; unresolvable/missing class → falls back to the default (both `true` and `false`); missing `job` key → default; a modern `data.command` payload behaves exactly as before; and an end-to-end regression firing `JobProcessing` for a legacy payload no longer throws.

### DIFF
--- a/src/Actions/MakeQueueTenantAwareAction.php
+++ b/src/Actions/MakeQueueTenantAwareAction.php
@@ -50,6 +50,15 @@ class MakeQueueTenantAwareAction
     {
         $payload = $this->getEventPayload($event);
 
+        /**
+         * Legacy string-job payloads ({"job":"Class@method","data":{...}}) - pushed by
+         * something other than Laravel's dispatcher, e.g. AWS EventBridge - have no
+         * serialized command. Resolve the class name straight from the payload instead.
+         */
+        if (! isset($payload['data']['command'])) {
+            return $this->resolveTenantAwarenessForJob($this->jobClassFromLegacyPayload($payload));
+        }
+
         $serializedCommand = $payload['data']['command'];
 
         if (! str_starts_with($serializedCommand, 'O:')) {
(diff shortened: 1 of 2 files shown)


masterix21:
Thanks for the excellent PR! Clear write-up, clean implementation, and great test coverage — including the regression test for modern payloads. Reviewed and tested locally, all green. Merging — much appreciated! 🙏

PR #641, opened 2026-07-17
docs: fix tenant-scoped job dispatch examples and Schedule::run() typo

AlirezaManiac:
## Summary
- The two `execute()`/`callback()` examples on this page dispatch a job as the return value of an arrow function (`fn () => dispatch(Job::class))`). `dispatch()`/`Job::dispatch()` return a lazy `PendingDispatch` — the actual queue push happens in its `__destruct()`. Returning that object out of the tenant closure keeps it alive past the point where `execute()`'s `finally` block has already reverted the current tenant, so the job is queued with the wrong tenant (or none), throwing `CurrentTenantCouldNotBeDeterminedInTenantAwareJob` on the worker.
- Rewrote both examples to dispatch as a bare statement inside a block closure (`function () { Job::dispatch(); }`), and added a warning callout explaining why the arrow-function form is unsafe here.
- Fixed `$schedule->run(...)` → `$schedule->call(...)` in the scheduler example — `Schedule::run()` doesn't exist on `Illuminate\Console\Scheduling\Schedule`.

## Test plan
- Docs-only change; no code paths affected.
- Verified against a real-world repro: the arrow-function form reliably threw `CurrentTenantCouldNotBeDeterminedInTenantAwareJob` when scheduling a tenant-scoped job dispatch; switching to a block closure fixed it.

### DIFF
--- a/docs/advanced-usage/executing-code-for-tenants-and-landlords.md
+++ b/docs/advanced-usage/executing-code-for-tenants-and-landlords.md
@@ -25,11 +25,16 @@ Here's another example, where a job is dispatched from a landlord API route:
 
 ```php
 Route::post('/api/{tenant}/reminder', function (Tenant $tenant) {
-    return json_encode([
-        'data' => $tenant->execute(fn () => dispatch(ExpirationReminder())),
-    ]);
+    $tenant->execute(function () {
+        ExpirationReminder::dispatch();
+    });
+
+    return json_encode(['success' => true]);
 });
 ```
+
+> **Warning**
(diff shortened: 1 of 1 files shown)


masterix21:
Hi @AlirezaManiac, thank you so much for this contribution! 🙏

The `PendingDispatch` caveat you documented is a real and subtle footgun — I verified it against the package internals, and the explanation is spot on. The rewritten route example and the warning callout are great additions.

One thing before we can merge: the PR description mentions fixing `$schedule->run(...)` → `$schedule->call(...)`, but the diff doesn't actually touch the existing scheduler example — it still reads:

```php
$schedule->run($tenant->callback(fn() => cache()->flush()))->daily();
```

The new (correct) example using `$schedule->call(...)` is only *added* below it, so after merging, the page would show the broken `run()` example right next to the working `call()` one, which could be confusing for readers.

Could you update the existing example to use `call()` instead? At that point, the extra code block may not even be needed — the short note about dispatching as a statement would probably be enough.

Thanks again for taking the time to dig into this and document it so thoroughly!

AlirezaManiac:
Hi @masterix21, thank you for the thorough review 
The branch has been updated. Please let me know if there's anything else you'd like me to adjust.

freekmurze:
Thanks!

PR #643, opened 2026-08-07
Restore the tenant that was current before a job was processed

### LINKED ISSUE #639: Processing a NotTenantAware job on the sync driver permanently forgets the dispatcher's current tenant — intended behaviour or bug?
## Versions

- spatie/laravel-multitenancy: 4.1.3 (behaviour unchanged on `main` at time of writing)
- laravel/framework: 12.x
- PHP: 8.4

## Description

When `queues_are_tenant_aware_by_default` is enabled, `MakeQueueTenantAwareAction::bindOrForgetCurrentTenant()` calls `Tenant::forgetCurrent()` when a `NotTenantAware` job starts processing, and nothing ever restores the tenant that was current before the job.

On a queue worker this goes unnoticed, because every job re-binds its own tenant and nothing meaningful runs between jobs. But when a job executes **in-process** — the `sync` driver, or `dispatchSync()` on any app — the job runs inside the dispatching request, so the *dispatcher* loses its current tenant for the remainder of the request:

```php
$tenant->makeCurrent();

SomeNotTenantAwareJob::dispatchSync();

Tenant::current(); // null — the request continues with no tenant
```

Everything tenant-dependent after that point misbehaves. The loudest symptom is dispatching a `TenantAware` job afterwards, which throws `CurrentTenantCouldNotBeDeterminedInTenantAwareJob` because no tenant was current at dispatch time. Tenant-scoped queries in the same window fail silently, which is worse.

The same mechanism affects `TenantAware` jobs run via `dispatchSync()` for a *different* tenant than the request's: after the job, the request continues under the **job's** tenant instead of its own.

## This contradicts the documented behaviour

From `docs/basic-usage/making-queues-tenant-aware.md`:

> "On the other hand, a job that is not tenant aware **will make no modifications to the current tenant**, which may still be set from a previous job."

The current implementation does modify it — `forgetCurrent()` — with no restore. (I suspect the `forgetCurrent()` was added as hygiene so a previous job's tenant can't bleed *into* a not-tenant-aware job, which is good — but the docs' contract that the surrounding context's tenant survives was lost in the process.)

## Reproduction (failing test against this package's suite)

Verified failing against `main` (`032e74c`): the first expectation passes (the job itself correctly runs without a tenant), the second fails with `Tenant::checkCurrent()` returning `false`. The package's existing `QueueIsTenantAwareByDefaultTest` passes in the same environment.

Drop this into `tests/Feature/TenantAwareJobs/` — it reuses the existing `NotTenantAwareTestJob` test class:

```php
<?php

use Spatie\Multitenancy\Tests\Feature\TenantAwareJobs\TestClasses\NotTenantAwareTestJob;
use Spatie\Valuestore\Valuestore;

beforeEach(function () {
    config()->set('multitenancy.queues_are_tenant_aware_by_default', true);
    config()->set('queue.default', 'sync');

    $this->tenant = Tenant::factory()->create();
    $this->valuestore = Valuestore::make(tempFile('tenantAware.json'))->flush();
});

it('does not lose the dispatching context\'s tenant after a sync NotTenantAware job', function () {
    $this->tenant->makeCurrent();

    dispatch(new NotTenantAwareTestJob($this->valuestore));

    // The job itself correctly ran without a tenant:
    expect($this->valuestore->get('tenantId'))->toBeNull();

    // But the dispatching context's tenant should still be current — this fails:
    expect(Tenant::checkCurrent())->toBeTrue()
        ->and(Tenant::current()->id)->toBe($this->tenant->id);
});
```

## Who hits this in practice

- Any app running the `sync` driver (Laravel's fallback default when `QUEUE_CONNECTION` is unset)
- Any app calling `dispatchSync()` mid-request
- Standard Laravel test suites (`QUEUE_CONNECTION=sync` in phpunit.xml): a model observer that dispatches a `NotTenantAware` job silently drops the tenant for the rest of the test — this is how we found it. Two earlier "fixes" in our codebase (marking a job `NotTenantAware`, adding a `Queue::fake()`) turned out to be band-aids over this same root cause.

## The docs and an existing test disagree — which reflects the intent?

I want to flag this carefully rather than assert it's simply a bug, because there is an existing test asserting the current behaviour: in `TenantAwareJobThatCannotFindTenantTest`, "it will forget any current tenant when starting a not tenant aware job" dispatches on the sync driver and then asserts `assertNull(Tenant::current())` *after* the dispatch has returned.

Reading that test's comments ("Simulate a tenant being set from a previous queue job"), its intent looks like worker hygiene: a stale tenant from a previous job must not bleed *into* a not-tenant-aware job — which makes complete sense. But the final assertion observes the forgotten tenant *after* the dispatch returns, which on the sync driver means it also pins down the side effect described above: the dispatching context loses its tenant permanently. That post-dispatch state is where the test and the documented contract ("will make no modifications to the current tenant") seem to point in opposite directions.

So my question is: which one reflects the intended behaviour?

- If the **docs** are the intent (jobs run with the right tenant context, but processing a job makes no net modification to the dispatcher's current tenant), then the action needs a restore step and that one test assertion needs adjusting.
- If the **test** is the intent (the current tenant is genuinely meant to be gone after a not-tenant-aware job is processed in-process), then the docs sentence quoted above is misleading and apps using the sync driver or `dispatchSync()` need to re-bind their tenant manually — which would be worth documenting.

For what it's worth, I have a working fix for the first interpretation validated against this package's test suite (restore the previously-current tenant on `JobProcessed`/`JobExceptionOccurred`, preserving the bleed-in protection during the job itself — all existing tests pass apart from the post-dispatch assertion discussed above). I'd be happy to open a PR if you confirm that's the direction you want; equally happy if the answer is "docs need updating" instead.

freekmurze:
Fixes #639.

`MakeQueueTenantAwareAction` forgets or rebinds the current tenant when a job starts, but never puts back what was there before. On a worker that goes unnoticed, but on the `sync` driver (or `dispatchSync()`) the job runs inside the dispatching request, so the dispatcher permanently loses its tenant.

The action now captures the current tenant on `JobProcessing` and restores it on `JobProcessed`/`JobExceptionOccurred`. It's a stack, so nested sync dispatches restore per level. Jobs still run with the right tenant (or none at all), nothing bleeds into a not tenant aware job, and a processed job no longer leaves a stale tenant behind on a worker either.

```php
$tenant->makeCurrent();

SomeNotTenantAwareJob::dispatchSync(); // runs without a current tenant

Tenant::current(); // still $tenant, was null before
```

`it will forget any current tenant when starting a not tenant aware job` asserted the side effect after the dispatch returned. It now asserts the job itself ran without a tenant, which was the intent. Also updated the docs sentence in `making-queues-tenant-aware.md` that described the old contract.


freekmurze:
Pushed b9dd49e after checking the backwards compatibility of the first commit. Two problems, both now covered by tests:

**Listeners on `JobProcessed` saw the restored tenant instead of the job's tenant.** A `Queue::after()` callback doing tenant-scoped work used to run with the job's tenant current. Because the restore listener was registered at package boot, it ran before anything the application registered and the callback saw the restored tenant (usually none). The listener is now registered when the first job starts processing, so it sits behind application listeners. `Queue::after()` and `JobFailed` now observe exactly what they observed before this PR.

**The fix did nothing when multitenancy is started more than once**, which is what Octane does per request. Each start registered its own pair of listeners with its own captured tenant, and the last one to restore won. The captured tenants are now shared between instances, so popping in reverse registration order leaves the outermost tenant current.

With those in place the only behaviour that changes is the one this PR is about: code outside the job lifecycle keeps the tenant it had. Everything participating in processing a job is unaffected.

freekmurze:
Fixed the `queue:retry` leak too.

The tenant can't be restored right after `JobRetryRequested`, because `RetryCommand::retryJob()` runs `refreshRetryUntil()` after the event, which unserializes the payload again. For a job holding models on a tenant connection that needs the tenant still current. So the boundary is the end of the command: the tenant that was current when the first `JobRetryRequested` fired is restored on `CommandFinished`.

That makes `Artisan::call('queue:retry')` from a request safe, which is where this actually bites. Running `php artisan queue:retry` from the shell was never really affected since the process exits anyway.

One caveat worth knowing: Laravel does not dispatch console events under `runningUnitTests()` unless the test opts in with `Illuminate\Foundation\Testing\WithConsoleEvents`, so `TenantAwareJobRetryTest` now uses that trait. In an application's own test suite `$this->artisan('queue:retry')` will not restore the tenant unless it opts in the same way. Production is unaffected, both `php artisan` and `Artisan::call()` fire the events.

Docs updated. Suite is at 97 passing, and all 11 matrix jobs (PHP 8.2 to 8.5, Laravel 11 to 13) are green.

PR #648, opened 2026-09-17
Return SUCCESS from TenantAware when no tenants exist

### LINKED ISSUE #647: `TenantAware` returns `-1` when no tenants exist, which breaks deployment pipelines on fresh installs
# `TenantAware` returns `-1` when no tenants exist, which breaks deployment pipelines on fresh installs

**Package version:**  4.2.0
**Laravel version:** 12.68.0
**PHP version:** 8.4.20

## Description

The `TenantAware` trait exits with `-1` when the tenant query returns zero rows:

```php
if ($tenantQuery->count() === 0) {
    $this->error('No tenant(s) found.');
    return -1;
}
```

`-1` is outside the valid exit code range (0–255) and is mapped to `255` by Symfony Console / PHP. Any deployment tool that stops on a non-zero exit — Laravel Forge, Envoyer, most CI runners — treats this as a hard failure.

## Reproduction

1. Fresh installation with an empty landlord database
2. Deploy script runs `php artisan tenants:artisan "migrate --force"` (or any command using `TenantAware`)
3. Output:

   ```
   No tenant(s) found.
   => Deployment failed: An unexpected error occurred during deployment.
   ```

4. All subsequent deploy steps (config cache, queue restart, …) are skipped

## Why this is a problem

"No tenants exist yet" is a normal, expected state on a first deploy, not an error condition. It is semantically different from "the tenant explicitly requested via `--tenant=` does not exist", which arguably *should* fail.

The current behaviour makes it impossible to write a deploy script that works both on a fresh server and on an existing one without shell workarounds such as `|| true` — which then also swallows genuine failures from `handle()`.

## Suggested change

Only return a failure code when the user explicitly filtered by `--tenant` and nothing matched. When no filter was given and there are simply no tenants, emit a warning and return `SUCCESS`:

```php
if ($tenantQuery->count() === 0) {

    if (blank($tenants)) {
        $this->warn('No tenants found, skipping.');
        return self::SUCCESS;
    }

    $this->error('No tenant(s) found.');
    return self::FAILURE;
}
```

Alternatively, an opt-in option such as `--skip-if-none` / `--allow-empty` would preserve the current default while letting deploy scripts choose the lenient behaviour.

Happy to open a PR (including tests) for whichever approach you prefer.

webard:
Closes #647.

`TenantAware::execute()` returned `-1` whenever the tenant query came back empty. On a fresh install that is the normal state - there are no tenants until the first one is created, so a deploy step like `php artisan tenants:artisan "migrate --force"` failed the whole pipeline.


masterix21:
Thanks

